Issue #25, opened 2020-11-30
Feature request: add aws-iam-authentication binary


markha73:

markha73:
The viya4-deployment docker container needs to be built with the aws-iam-authenticator in order to deploy into an AWS cluster. The following commands need to be added to the docker file used to create the viya4-deployment docker container:

curl -o aws-iam-authenticator https://amazon-eks.s3.us-west-2.amazonaws.com/1.18.9/2020-11-02/bin/linux/amd64/aws-iam-authenticator 
chmod 755 aws-iam-authenticator 
mkdir ~/bin 
mv aws-iam-authenticator ~/bin 
ls -aslrt ~/bin
export PATH="$PATH:$HOME/bin"

PR #28, opened 2020-12-07
add aws-iam-authenticator

enderm:
to address https://github.com/sassoftware/viya4-deployment/issues/25


NormJohnIV:
@enderm: I updated the commit to be more inline with the other binaries. Please review

enderm:
lgtm

PR #53, opened 2021-01-28
Metrics server

NormJohnIV:
deploy metrics server on clusters that don't have it installed already.


roles/metrics-server/defaults/main.yml
```diff
@@ -0,0 +1,7 @@
+metrics_server_name: metrics-server
+kubeconfig: "~/.kube/config"
```
thpang, line 2: Is this supplied by the user? Wondering why this value is hardcoded here.

NormJohnIV: That is a default value. Its actually set to the one the user provide or we read from the tfstate. See final var set here: https://github.com/sassoftware/viya4-deployment/blob/main/playbooks/playbook.yaml#L83 . This is how we do it in all the roles. At a future day we may simplify but want to stay consistent for now.

thpang: cool, cool.

PR #78, opened 2021-04-14
Nfs subdir

NormJohnIV:
- uninstall efs-provsioner
- uninstall nfs-client-provisioner
- install nfs-subdir-external-provisioner


enderm (commented):
should "prviate" in multiple var names rather be "private"?
It is consistent, so it's not a functional issue.

PR #96, opened 2021-05-26
Add vars for configuring 3rd party helm chart

NormJohnIV:
Add vars to config settings for the 3rd party helm charts

1. Repo url
2. Repo version
3. Helm values
4. Helm release name
5. Deployment namespace (except for metrics-server which always deploys in kube-system.)

Also, bumped cert-manger version to 1.3.0


roles/common/tasks/main.yaml
```diff
@@ -19,13 +19,17 @@
     - name: Parse tfstate
       set_fact: 
         tfstate: "{{ lookup('tfstate', TFSTATE) }}"
+    - name: default LOADBALANCER_SOURCE_RANGES
+      set_fact:
+        LOADBALANCER_SOURCE_RANGES: ["0.0.0.0/0"]
```
thpang, line 24: Is there a way to have a default that's more secure? Wondering about the effects of having the default being wide open.

thpang (changes requested):
Added a couple of comments. This looks good. Much cleaner and compact.

roles/common/tasks/main.yaml
NormJohnIV: Not really. If the user does not provide LOADBALANCER_SOURCE_RANGES then the deployment would not be reachable at all. So, we default to 0.0.0.0/0. This is essentially what the same as what the tools do when no loadBalancerSourceRanges are set.

thpang: Thanks for the info. That's what I needed to know.

docs/CONFIG-VARS.md
```diff
@@ -215,15 +216,42 @@ When setting V4_CFG_TLS_MODE to a value other than "disabled" and no V4_CFG_TLS_
...
+| | | | | | | |
+| NFS_CLIENT_NAMESPACE | nfs-subdir-external-provisioner helm install namespace | string | nfs-client | false | | baseline |
+| NFS_CLIENT_CHART_URL | nfs-subdir-external-provisioner helm chart url | string | https://kubernetes-sigs.github.io/nfs-subdir-external-provisioner/ | false | | baseline |
+| NFS_CLIENT_CHART_NAME | nfs-subdir-external-provisioner helm chart name | string | nfs-subdir-external-provisioner | false | | baseline |
+| NFS_CLIENT_CHART_VERSION | nfs-subdir-external-provisioner helm chart version | string | 4.0.8| false | | baseline |
+| NFS_CLIENT_CONFIG | nfs-subdir-external-provisioner helm values | string | see [here](../roles/baseline/defaults/main.yml) | false | | baseline |
+| | | | | | | |
+| METRICS_SERVER_CHART_URL | metrics-server helm chart url | string | https://charts.bitnami.com/bitnami/ | false | | baseline |
+| METRICS_SERVER_CHART_NAME | metrics-server helm chart name | string | metrics-server | false | | baseline |
+| METRICS_SERVER_CHART_VERSION | metrics-server helm chart version | string | 5.3.5 | false | | baseline |
```
hahewlet, line 256: This one seems a bit misleading. Two of our 3 cloud providers currently install a different version of the metric server, which is what we use. Can  can we convey this here?

thpang:
Only AWS requires us to install the metric server Azure/GCP already install that item. We don't know the current version that they are installing. But we can control the AWS item.
________________________________

docs/CONFIG-VARS.md
```diff
@@ -215,15 +216,42 @@ When setting V4_CFG_TLS_MODE to a value other than "disabled" and no V4_CFG_TLS_
...
+| INGRESS_NGINX_CHART_VERSION | ingress-nginx helm chart version | string | 3.20.1| false | | baseline |
+| INGRESS_NGINX_CONFIG | ingress-nginx helm values | string | see [here](../roles/baseline/defaults/main.yml) | false | | baseline |
+| | | | | | | |
+| NFS_CLIENT_NAMESPACE | nfs-subdir-external-provisioner helm install namespace | string | nfs-client | false | | baseline |
+| NFS_CLIENT_CHART_URL | nfs-subdir-external-provisioner helm chart url | string | https://kubernetes-sigs.github.io/nfs-subdir-external-provisioner/ | false | | baseline |
+| NFS_CLIENT_CHART_NAME | nfs-subdir-external-provisioner helm chart name | string | nfs-subdir-external-provisioner | false | | baseline |
+| NFS_CLIENT_CHART_VERSION | nfs-subdir-external-provisioner helm chart version | string | 4.0.8| false | | baseline |
+| NFS_CLIENT_CONFIG | nfs-subdir-external-provisioner helm values | string | see [here](../roles/baseline/defaults/main.yml) | false | | baseline |
+| | | | | | | |
+| METRICS_SERVER_CHART_URL | metrics-server helm chart url | string | https://charts.bitnami.com/bitnami/ | false | | baseline |
```
hahewlet, line 254: If an existing metric-server is installed, these options will be ignored. Note Managed Amazon Elastic Kubernetes Service is the only supported provider that does not include a metric-server by default.

thpang: That's not true, GCP also has a metric server active in your cluster. Currently AWS is the only provider NOT installing metric server by default.

NormJohnIV: updated

PR #101, opened 2021-06-08
add ingress.class annotation for nginx

enderm:


enderm:
fixed via https://github.com/sassoftware/viya4-deployment/commit/472d92f72bcd5bfe85922a75421fbe9ff042e158

PR #103, opened 2021-06-08
parse file as yaml

NormJohnIV:


jkmin96:
I tested this and I verified that main kustomization.yaml was created correctly.

jkmin96 (commented):
I tested this and I verified that main kustomization.yaml was created correctly.
PVCs got created from the resources.yaml w/o problem.

PR #105, opened 2021-06-21
Basic contour install

NormJohnIV:
Add basic support for installing contour during baseline. This will not configure viya to use contour. It is install only


thpang:
LGTM

PR #107, opened 2021-06-21
add elastic folder check

NormJohnIV:
Added additional folder check to ensure elasticsearch is not enable when the order does not include elasticsearch.


thpang:
LGTM

hahewlet:
Talked to @NormJohnIV. signing off on testing.

PR #112, opened 2021-06-24
Enable cluster-autoscaler on AWS

NormJohnIV:
Support deploying cluster autoscaler in aws


hahewlet:
Need a reference in the readme to inheriting CLUSTER_AUTOSCALER_ACCOUNT from terrform.tfstate

hahewlet:
@NormJohnIV  the default region for the K8s cluster-autoscaler is us-east-1 as defined in https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/values.yaml. Can viya4-deployment change that to match whatever region the cluster resides in?. For performance purposes that would been a good thing.
    
yes I can override that with CLUSTER_AUTOSCALER_LOCATION. Acting as the customer, I'd think hey I told viya4-iac-aws what region I wanted. I told viya4-deployment to use my terraform.tfstate file, so I expect my region choice to be honored.

hahewlet:
Testing completed signing off.

PR #115, opened 2021-07-11
2020.1.3 tls components support

NormJohnIV:
- add support for components
- add support for 2020.1.3+ tls overlays

### DIFF
--- a/roles/vdm/library/siteconfig_info.py
+++ b/roles/vdm/library/siteconfig_info.py
@@ -2,45 +2,28 @@ from ansible.module_utils.basic import *
 import glob
 import yaml
 import os
+from enum import Enum, auto
+
+class Overlay(Enum):
+  COMPONENT = "components"
+  CONFIGURATION = "configurations"
+  GENERATOR = "generators"
+  RESOURCE = "resources"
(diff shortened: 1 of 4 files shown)


roles/vdm/library/siteconfig_info.py
```diff
@@ -2,38 +2,28 @@
 import glob
 import yaml
 import os
+from enum import Enum, auto
+
+class Overlay(Enum):
+  COMPONENT = "components"
```
thpang, line 8: Just curious why are the defined items singular when the assigned values are plural?

roles/vdm/library/siteconfig_info.py
```diff
@@ -2,38 +2,28 @@
...
+  CONFIGURATION = "configurations"
+  GENERATOR = "generators"
+  RESOURCE = "resources"
+  TRANSFORMER = "transformers"
```
thpang, line 12: Just curious why are the defined items singular when the assigned values are plural?

hahewlet:
Testing is completed. Signing off.

agianni:
This should say 2021.1.3 at the top right?

PR #125, opened 2021-08-02
add support for data server operator

NormJohnIV:

PR #129, opened 2021-08-16
re-add viya-monitoring task

enderm:


NormJohnIV:
Fixed in https://github.com/sassoftware/viya4-deployment/pull/125

PR #134, opened 2021-08-30
fix front-door tls for pre-2021.1.3 cadence versions

enderm:
tested with 2021.1.2 (to verify fix) and with 2021.1.4 to make sure it's not breaking more recent version.


hahewlet:
Testing is completed. signing off.

Issue #133, opened 2021-08-30
RegEx Syntax issue in roles/vdm/defaults/main.yaml


nachai9:
I am running into same issue as issue # 87 when I am using Azure Container Registry. Can you pls reopen it for me.

Here is the sample error and attaching ansible-vars.yaml for reference.

ERROR! We were unable to read either as JSON nor YAML, these are the errors we got from each:
JSON: Expecting value: line 1 column 1 (char 0)

 

Syntax Error while loading YAML.
  found unknown escape character

 

The error appears to be in '/r/ge.unx.sas.com/vol/vol610/u61/nachai/viya4_3/viya4-deployment/roles/vdm/defaults/main.yaml': line 15, column 60, but may
be elsewhere in the file depending on the exact syntax problem.

 

The offending line appears to be:

 

V4_CFG_CR_URL: https://cr.sas.com
V4_CFG_CR_HOST: "{{ V4_CFG_CR_URL | regex_replace('^https?:\/\/(.*)\/?', '\\1') }}"
                                                           ^ here
We could be wrong, but this one looks like it might be an issue with
missing quotes. Always quote template expression brackets when they
start a value. For instance:

 

    with_items:
      - {{ foo }}

 

Should be written as:

 

    with_items:
      - "{{ foo }}"
[ansible-vars.zip](https://github.com/sassoftware/viya4-deployment/files/7076894/ansible-vars.zip)

AWSmith0216:
Just a thought, but your V4_CFG_CR_URL value doesn't match the regex. Perhaps try adjusting your value to 'V4_CFG_CR_URL: https://nachai3acr.azurecr.io' and see what happens.

nachai9:
I am getting following error with V4_CFG_CR_URL:

Monday 30 August 2021  10:08:13 -0400 (0:00:01.443)       0:00:21.601 *********
ERROR! We were unable to read either as JSON nor YAML, these are the errors we got from each:
JSON: Expecting value: line 1 column 1 (char 0)

Syntax Error while loading YAML.
  found unknown escape character

The error appears to be in '/r/ge.unx.sas.com/vol/vol610/u61/nachai/viya4_3/viya4-deployment/roles/vdm/defaults/main.yaml': line 15, column 60, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

V4_CFG_CR_URL: https://cr.sas.com
[...]

PR #136, opened 2021-08-31
switch quoting in variable definition

enderm:
fixes issue #133


thpang:
@enderm what testing has been done on this one? Asking to see if it's been tested against aws/azure/gcp at this point.

enderm:
no testing except my own (local env vs clglgn), and verification by one of the folks who ran into it.
I don't see how the different clouds would matter here. The code in question is executed locally.

hahewlet:
Testing is completed. Signing off .

PR #144, opened 2021-09-10
add cadence control to sizing transformers

enderm:
- overlays/cas-server/require-cas-label.yaml >= 2021.1.2
- overlays/scaling/single-replica/transformer.yaml >= 2021.1.1


thpang:
Based on conversation and requirements. LGTM.

PR #146, opened 2021-09-15
Update ansible-vars-iac.yaml to include Postgres

GonzaloUlla:
Per [CONFIG-VARS.md#postgres](https://github.com/sassoftware/viya4-deployment/blob/main/docs/CONFIG-VARS.md#postgres), `V4_CFG_POSTGRES_SERVERS` is required. 

Adding an `internal: true` example configuration to `ansible-vars-iac.yaml`

### DIFF
--- a/examples/ansible-vars-iac.yaml
+++ b/examples/ansible-vars-iac.yaml
@@ -23,6 +23,9 @@ V4_CFG_INGRESS_FQDN: <desired_fqdn>
 V4_CFG_TLS_MODE: "full-stack" # [full-stack|front-door|disabled]
 
 ## Postgres
+V4_CFG_POSTGRES_SERVERS:
+  default:
+    internal: true
 
 ## LDAP
 V4_CFG_EMBEDDED_LDAP_ENABLE: true


thpang:
That is incorrect. The value is pulled from the IAC terraform.tfstate file here, There should be no value. If it's passed from IAC it'll use the external DB. If it's not provided, then it'll be crunchy that gets setup.

GonzaloUlla:
> If it's not provided, then it'll be crunchy that gets setup.

Actually, if Postgres `default.internal` key is not provided, the Ansible task fails.

Check issue #147

thpang:
I see your point here; however, if you've chosen not to setup an external database the docs do say you need the entry. It's just they, or in this case you, have not added that entry. Having the default value you suggest is not always correct.

GonzaloUlla:
Whether you choose to setup an external database or not, you always need to set the `default` and `default.internal` elements, per [CONFIG-VARS.md#postgres](https://github.com/sassoftware/viya4-deployment/blob/main/docs/CONFIG-VARS.md#postgres)

If those entries were not provided, why not assume it is intended to use crunchy instead of failing?

Leaving an empty line after `## Postgres` in [ansible-vars-iac.yaml#L25](https://github.com/sassoftware/viya4-deployment/blob/main/examples/ansible-vars-iac.yaml#L25) and not defining a default behavior, it's error-prone. Example files should work out of the box IMHO.

PR #148, opened 2021-09-16
add INGRESS_MODE suppport

enderm:
tested with all three cloud "provider" values, both with tfstate and directly. Verified that the correct annotations got set on the ingress controller service.

Using branch "infra_mode". The existing branch "private_cluster" is stale and should be deleted. Leaving it around for now in case reviewers want to reference it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,7 @@ This following information is parsed from the integration:
   - V4_CFG_CLUSTER_NODE_POOL_MODE
   - CLUSTER_AUTOSCALER_ACCOUNT
   - CLUSTER_AUTOSCALER_LOCATION
+  - INFRA_MODE
 
 ### Customize Deployment Overlays
 
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ This following information is parsed from the integration:
   - CLUSTER_AUTOSCALER_ACCOUNT
   - CLUSTER_AUTOSCALER_LOCATION
 - Ingress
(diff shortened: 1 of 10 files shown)


thpang (commented):
Looks good but need to wait on changes to viya4-iac-azure. Also need to ensure that the AWS/GCP iac items are not broken with these changes.

enderm:
not seeing anything in the code that would make this backward-incompatible. Still, I did a successful deployment on a gke cluster successfully with this branch, so I'm quite confident we are not breaking aws or gcp.

thpang:
Looks good waiting on viya4-iac-azure PR Number 120.

README.md
```diff
@@ -138,6 +138,8 @@ This following information is parsed from the integration:
   - V4_CFG_CLUSTER_NODE_POOL_MODE
   - CLUSTER_AUTOSCALER_ACCOUNT
   - CLUSTER_AUTOSCALER_LOCATION
+- Ingress
+  - default value for V4_CFG_INGRESS_MODE (from CLUSTER_API_MODE)
```
thpang, line 142: I think just having `V4_CFG_INGRESS_MODE` would suffice here. Minor change. Like `V4_CFG_POSTGRES_SERVERS` on line 135 above.

It would look like this:

```yaml
- Ingress
    - V4_CFG_INGRESS_MODE (from CLUSTER_API_MODE)
```

thpang (changes requested):
Small change in doc.

thpang (approved):
LGTM

PR #149, opened 2021-09-23
add selector to exclude storageClassName patch from some StatefulSets 

### LINKED ISSUE #121: "kustomize - Generate deployment manifest" task fails in orders containing SAS Workload Management
When an order contains SAS Workload Management, the playbook fails when generating the deployment manifest.

The issue occurs because sas-storageclass.yaml (https://github.com/sassoftware/viya4-deployment/blob/7e89b16edc7369683422b1950dd57a261be320b3/roles/vdm/templates/transformers/sas-storageclass.yaml) attempts to add a Storage Class name to the volumeClaimTemplate for every Stateful Set.

However the sas-workload-orchestrator Stateful Set that comes with SAS Workload Management does not contain a volumeClaimTemplate attribute, leading to a build error;
```
Error: failed to apply json patch '- op: add
  path: /spec/volumeClaimTemplates/0/spec/storageClassName
  value: sas': add operation does not apply: doc is missing path: "/spec/volumeClaimTemplates/0/spec/storageClassName": missing value
```
A possible workaround, that I have tested, would be to explicitly exclude sas-workload-orchestrator using an annotation selector;

```yaml
apiVersion: builtin
kind: PatchTransformer
metadata:
  name: sas-ss-storageclass
patch: |-
  - op: add
    path: /spec/volumeClaimTemplates/0/spec/storageClassName
    value: sas
target:
  group: apps
  kind: StatefulSet
  annotationSelector: sas.com/component-name != sas-workload-orchestrator
```

### LINKED ISSUE #126: "kustomize - Generate deployment manifest" task fails in orders containing SAS Model Risk Management
When I tried to deploy an order which including **SAS Model Risk Management**, I got below error:

```
"msg": "non-zero return code",
    "rc": 1,
    "start": "2021-08-09 12:13:23.185705",
    "stderr": "Error: failed to apply json patch '- op: add\n  path: /spec/volumeClaimTemplates/0/spec/storageClassName\n  value: sas': add operation does not apply: doc is missing path: \"/spec/volumeClaimTemplates/0/spec/storageClassName\": missing value",
    "stderr_lines": [
        "Error: failed to apply json patch '- op: add",
        "  path: /spec/volumeClaimTemplates/0/spec/storageClassName",
        "  value: sas': add operation does not apply: doc is missing path: \"/spec/volumeClaimTemplates/0/spec/storageClassName\": missing value"
    ],
```

When I checked opensource K8S deployment including SAS Model Risk Management, I see this new statefulset was created:

```
bash-4.4$ k -n trs get statefulsets | grep risk
sas-risk-cirrus-search      1/1     33h
```

And these PVCs were added:

```
sas-risk-cirrus-builder                               Bound    pvc-174a58a0-f186-486c-b581-ae704fd3ae3e   2Gi        RWX            nfs-client     33h
sas-risk-cirrus-search-data                           Bound    pvc-978d768f-aac0-44f7-9d18-d342e62df812   2Gi        RWX            nfs-client     33h
```

enderm:
make sure that StatefulSets that have no volumeClaimTemplates do not get the sas-ss-storageclass overlay.

fixes #126
fixes #121 

tested by successfully deploying order with both Risk Management and SWO.

### DIFF
--- a/roles/vdm/templates/transformers/sas-storageclass.yaml
+++ b/roles/vdm/templates/transformers/sas-storageclass.yaml
@@ -23,4 +23,4 @@ patch: |-
 target:
   group: apps
   kind: StatefulSet
-  annotationSelector: sas.com/component-name notin (sas-risk-cirrus-search sas-workload-orchestrator)
+  annotationSelector: sas.com/component-name notin (sas-risk-cirrus-search,sas-workload-orchestrator)
--- a/roles/vdm/templates/transformers/sas-storageclass.yaml
+++ b/roles/vdm/templates/transformers/sas-storageclass.yaml
@@ -23,3 +23,4 @@ patch: |-
 target:
   group: apps
   kind: StatefulSet
+  annotationSelector: sas.com/component-name notin (sas-risk-cirrus-search sas-workload-orchestrator)


jkmin96 (approved):
I pulled storageclass branch, built a docker image and run Viya4 deployment with an order including MRM and SWO.
Deployment was done successfully. Pods are up and running well.

thpang:
Could the transformers for these products not do this? Not a good idea to start hard coding product information into the deployment code base. @enderm work with the product teams to add transforms to their products to handle this and make sure they implement the same way. I am sure others will also have these issues.

enderm:
@Thomas ***@***.***> I agree that encoding production information is not ideal.

Can you please explain what you have in mind? How could the transformers for those product achieve what we need here?

thpang:
If the development teams cannot create an empty transformer for the storage class, they will need to document how to add the entry into the sas-storageclass.yaml. If the teams are choosing other options for storage class type, this needs to be addressed. Granted I know the code works here; however, the list will soon grow. Going to bring this up to the K8s sig and see what their thoughts are on this one. Need a solution not a patch.

jkmin96:
Are you done with this?
If yes, could you please create a private docker image which including this fix?

thpang:
Merging to keep folks moving will re-address in the future for a more robust solution.

PR #150, opened 2021-10-07
Updating `network` to support new directory formats for v1 kubernetes change

thpang:


thpang:
@enderm if you could review these changes for the `ingres` updates for 2021.1.6. Need to create a release for this for folks to use. Thx.

Issue #156, opened 2021-10-20
/site-config/vdm/security is not created if one does not include a custom CA Cert.


thpang:
Need to create this directory given a customer may not have a custom CA cert, but does has their own .crt and .key file.

PR #157, opened 2021-10-20
Creating security dir if needed

thpang:
Fixes: #156


thpang:
User only provided the .key and .crt files no CA for this round. So the /security directory needs to be there. @manoatsas  and @enderm if you can review.

manoatsas:
I'm not familiar with the intricacies of this project to be able to comment one way or the other. You can proceed to merging with other approval.

PR #160, opened 2021-11-01
add cluster-api label support

enderm:
- tested against current RTR
- verified that the block does not wait for the timeout if no pods exist

### DIFF
--- a/roles/vdm/tasks/deploy.yaml
+++ b/roles/vdm/tasks/deploy.yaml
@@ -11,6 +11,19 @@
     - install
     - update
 
+- name: prereqs - cluster-api
+  shell: |
+    kubectl --kubeconfig {{ KUBECONFIG }} apply -n {{ NAMESPACE }} --selector="sas.com/admin=cluster-api" -f {{ DEPLOY_DIR }}/site.yaml
+    kubectl --kubeconfig {{ KUBECONFIG }} wait --for condition=established --timeout=60s -l "sas.com/admin=cluster-api" crd
+  register: result
+  failed_when:
+    - result["stderr"]|length > 0
+    - result["stderr"] is not regex(".* no matches for kind .* in version .*")  
+    - result["stderr"] is not regex(".*Warning.*")
+  tags:
+    - install
+    - update
+
 - name: prereqs - cluster-wide
   shell: |
     kubectl --kubeconfig {{ KUBECONFIG }} apply -n {{ NAMESPACE }} --selector="sas.com/admin=cluster-wide" -f {{ DEPLOY_DIR }}/site.yaml


thpang (commented):
Any issues in testing a deployment where this flag was applied?

enderm:
I tested viya deployment with the current RTR, which do not have this new flag yet. So it was basically a regression test. 
I have not tested any deployments that have the new flag (not sure that it is possible at this point, short of applying overlays).
The straightforward nature of the change makes that seem a reasonable approach.

AWSmith0216:
I'm surprised this would have worked as-is. If you do a wait for a label that doesn't exist, that should fail:

```
hostname> kc wait --for condition=established --timeout=60s -l "sas.com/admin=foo" crd
error: no matching resources found
hostname> echo $?
1
```

Are you sure you didn't already have CRDs with this label on your cluster?

thpang:
There would need to be another conditional that looked for this : error: no matching resources found

```yaml
    - result["stderr"] is not regex(".* error: no matching resources found .*")  
```

Or something similar. I have not tested.

enderm:
I reran this, after the recent comments.
Here is the full stderr from the new "cluster-api" block:
```
    "stderr_lines": [
        "unable to recognize \"/data/eurmae-tf-aks/viya/site.yaml\": no matches for kind \"CASDeployment\" in version \"viya.sas.com/v1alpha1\"",
        "unable to recognize \"/data/eurmae-tf-aks/viya/site.yaml\": no matches for kind \"Pgcluster\" in version \"webinfdsvr.sas.com/v1\"",
        "error: no matching resources found"
    ],
```
    
 and here from the "cluster-wide" block:
 
```
    "stderr_lines": [
        "Warning: apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition",
        "unable to recognize \"/data/eurmae-tf-aks/viya/site.yaml\": no matches for kind \"CASDeployment\" in version \"viya.sas.com/v1alpha1\"",
        "unable to recognize \"/data/eurmae-tf-aks/viya/site.yaml\": no matches for kind \"Pgcluster\" in version \"webinfdsvr.sas.com/v1\""
    ],  
```

So, it will work in both cases, because the "no matches for kind" is being found. 
But I'm not sure that will always be the case. 
I think adding an explicit condition to "failed_when"  for "no matching resources found" - to both blocks - is definitely indicated here. I'll open another PR.

AWSmith0216:
I know you've already fixed this, but just as an FYI, the messages similar to:

`"unable to recognize \"/data/eurmae-tf-aks/viya/site.yaml\": no matches for kind \"CASDeployment\" in version \"viya.sas.com/v1alpha1\"",`

...will only show up on the first deployment to a cluster. On subsequent deployments I don't believe you'll see that, so the fix you made in your other PR would be needed.

PR #162, opened 2021-11-02
add condition to sas.com/admin label blocks

enderm:
- add condition to prevent failure when no matching resources are found


thpang:
LGTM

PR #167, opened 2021-11-15
Adjusting failure check for deploy_logging_open.sh script

thpang:
Working on removing failure guard.


thpang:
Tested and verified.

thpang:
Verified and tested.

thpang:
Pushing to support a breaking change from the logging and monitoring team.

PR #191, opened 2022-02-28
(IAC-418) Update ingress-only TLS Mode Task

jarpat:
changes: added a check for cadence, if below 2021.2.4 & using  ingress-only will quit the playbook early with a meaningful error message.
doc: Add note that ingress-only TLS mode is supported for cadence versions 2021.2.4+

testing:
Performed deployments with ingress-only TLS for various cadences to make sure the task was performing correctly. GCP 1.22.6-gke.300 cluster 
stable: 2021.2.3: task correctly quits the playbook early
stable: 2021.2.4: ingress-only is accepted, the deployment starts
lts: 2021.2: task correctly quits the playbook early
fast: ingress-only is accepted, the deployment starts


jarpat:
Will also be adding an additional task to check for cadence if ingress-only is set. Will exit early with a message if not compatible.

thpang:
LGTM and can be merged when JK finishing his next round of testing.

jarpat:
merging this into staging to create a test image

PR #194, opened 2022-03-07
(IAC-445) (IAC-444) Update Ingress-ngnix version and apply CVE Mitigation

jarpat:
## Changes

- (IAC-444) This change updates the default ingress-nginx installed on K8s clusters whose version is <= 1.21.X to version 3.40.0 (appVersion 0.50.0)
- (IAC-445) Also included in this PR is a task that will apply a patch to the ingress-nginx helm chart to mitigate [CVE-2021-25742](https://nvd.nist.gov/vuln/detail/CVE-2021-25742). This patch will automatically be applied to all deployments on clusters whose  version is >= 1.22.X

### Testing For IAC-444:

1. GKE 1.22.4-gke.1501 Cluster
Ran viya4-deployment with a fast:2020 order, all tasks set to install without specifying INGRESS_NGINX_CHART_VERSION (baseline,viya,cluster-logging,cluster-monitoring,viya-monitoring,install) and waited for Viya, Kibana, & Grafana to stabilize and become usable
ingress-nginx appVersion 1.1.0 was installed. 

2.  GKE 1.21.6-gke.1503 Cluster
Ran viya4-deployment with a fast:2020 order, all tasks set to install without specifying INGRESS_NGINX_CHART_VERSION (baseline,viya,cluster-logging,cluster-monitoring,viya-monitoring,install) and waited for Viya, Kibana, & Grafana to stabilize and become usable
ingress-nginx appVersion 0.50.0 was installed. 

### Testing for IAC-445

1.  GKE 1.22.4-gke.1501 Cluster
Ran viya4-deployment with all tasks set to install without specifying INGRESS_NGINX_CHART_VERSION (baseline,viya,cluster-logging,cluster-monitoring,viya-monitoring,install) and waited for Viya, Kibana, & Grafana to stabilize and become usable
ingress-nginx appVersion 1.1.0 was installed.
The ingress-nginx-controller configmap was present with required configuration data for the mitigation 

```yaml
data:
  allow-snippet-annotations: "true"
  annotation-value-word-blocklist: load_module,lua_package,_by_lua,location,root,proxy_pass,serviceaccount,{,},\
  large-client-header-buffers: 4 32k
  use-forwarded-headers: "true"
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: ingress-nginx
    meta.helm.sh/release-namespace: ingress-nginx
  creationTimestamp: "2022-03-07T16:22:09Z"
  labels:
    app.kubernetes.io/component: controller
    app.kubernetes.io/instance: ingress-nginx
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: ingress-nginx
    app.kubernetes.io/version: 1.1.0
    helm.sh/chart: ingress-nginx-4.0.13
  name: ingress-nginx-controller
  namespace: ingress-nginx
  resourceVersion: "26329"
  uid: cb1660de-fbce-42cf-bb8c-8e4d2410b3a0
```
  
2. GKE 1.21.6-gke.1503 Cluster
Ran viya4-deployment with all tasks set to install without specifying INGRESS_NGINX_CHART_VERSION (baseline,viya,cluster-logging,cluster-monitoring,viya-monitoring,install) and waited for Viya, Kibana, & Grafana to stabilize and become usable
Verified that the patch was not applied.
ingress-nginx appVersion 0.50.0 was installed. 

### Re-Testing for IAC-445 (After Latest Updates)
1.  GKE 1.22.4-gke.1501 Cluster
Ran viya4-deployment with all tasks set to install without specifying INGRESS_NGINX_CHART_VERSION (baseline,viya,cluster-logging,cluster-monitoring,viya-monitoring,install) and waited for Viya, Kibana, & Grafana to stabilize and become usable
ingress-nginx appVersion 1.1.0 was installed.
The ingress-nginx-controller configmap was present with required configuration data for the mitigation 

  
2. EKS 1.21 Cluster
Ran viya4-deployment with all tasks set to install without specifying INGRESS_NGINX_CHART_VERSION (baseline,viya,cluster-logging,cluster-monitoring,viya-monitoring,install) and waited for Viya, Kibana, & Grafana to stabilize and become usable
ingress-nginx appVersion 0.50.0 was installed. 
The ingress-nginx-controller configmap was present with required configuration data for the mitigation

### DIFF
--- a/roles/baseline/defaults/main.yml
+++ b/roles/baseline/defaults/main.yml
@@ -54,6 +54,9 @@ INGRESS_NGINX_CONFIG:
 
     config:
       use-forwarded-headers: "true"
+      allow-snippet-annotations: "true"
+      large-client-header-buffers: "4 32k"
+      annotation-value-word-blocklist: "load_module,lua_package,_by_lua,location,root,proxy_pass,serviceaccount,{,},\\"
     tcp: {}
     udp: {}
     lifecycle:
@@ -61,16 +64,6 @@ INGRESS_NGINX_CONFIG:
         exec:
           command: ["/bin/sh", "-c", "sleep 5; /usr/local/nginx/sbin/nginx -c /etc/nginx/nginx.conf -s quit; while pgrep -x nginx; do sleep 1; done"]
     terminationGracePeriodSeconds: 600
-
-# Ingress-nginx - CVE-2021-25742 Mitigation
-INGRESS_NGINX_CVE_PATCH:
-  controller:
-    config:
-      allow-snippet-annotations: "true"
-      large-client-header-buffers: "4 32k"
-      use-forwarded-headers: "true"
-      annotation-value-word-blocklist: "load_module,lua_package,_by_lua,location,root,proxy_pass,serviceaccount,{,},\\"
-
 
 ## Nfs-subdir-external-provisioner
 NFS_CLIENT_NAME: nfs-subdir-external-provisioner
--- a/roles/baseline/defaults/main.yml
+++ b/roles/baseline/defaults/main.yml
@@ -30,8 +30,8 @@ ingressVersions:
   k8sMinorVersionCeiling:
     value: 21
     api:
-      chartVersion: 3.20.1
-      appVersion: 0.43.0
+      chartVersion: 3.40.0
+      appVersion: 0.50.0
   k8sMinorVersionFloor:
     value: 22
     api:
@@ -61,6 +61,16 @@ INGRESS_NGINX_CONFIG:
         exec:
           command: ["/bin/sh", "-c", "sleep 5; /usr/local/nginx/sbin/nginx -c /etc/nginx/nginx.conf -s quit; while pgrep -x nginx; do sleep 1; done"]
     terminationGracePeriodSeconds: 600
+
+# Ingress-nginx - CVE-2021-25742 Mitigation
+INGRESS_NGINX_CVE_PATCH:
+  controller:
+    config:
+      allow-snippet-annotations: "true"
+      large-client-header-buffers: "4 32k"
+      use-forwarded-headers: "true"
+      annotation-value-word-blocklist: "load_module,lua_package,_by_lua,location,root,proxy_pass,serviceaccount,{,},\\"
+
 
 ## Nfs-subdir-external-provisioner
 NFS_CLIENT_NAME: nfs-subdir-external-provisioner
--- a/docs/CONFIG-VARS.md
+++ b/docs/CONFIG-VARS.md
@@ -306,8 +306,8 @@ Cluster-autoscaler is currently only used for AWS EKS clusters. GCP GKE and Azur
 | INGRESS_NGINX_NAMESPACE | ingress-nginx helm install namespace | string | ingress-nginx | false | | baseline |
 | INGRESS_NGINX_CHART_URL | ingress-nginx helm chart url | string | https://kubernetes.github.io/ingress-nginx | false | | baseline |
 | INGRESS_NGINX_CHART_NAME | ingress-nginx helm chart name | string | ingress-nginx | false | | baseline |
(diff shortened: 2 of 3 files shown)


roles/baseline/defaults/main.yml
```diff
@@ -62,6 +62,16 @@ INGRESS_NGINX_CONFIG:
...
+# Ingress-nginx - CVE-2021-25742 Mitigation
+INGRESS_NGINX_CVE_PATCH:
+  controller:
+    config:
+      allow-snippet-annotations: "true"
+      large-client-header-buffers: "4 32k"
+      use-forwarded-headers: "true"
+      annotation-value-word-blocklist: "load_module,lua_package,_by_lua,location,root,proxy_pass,serviceaccount,{,},\\"
+
+
```
thpang, line 74: This is our new default, why is this not included in line 48 above? I don't see why do we have this? It's novel but don't believe the abstraction is worth the cost in support. Thoughts?

thpang (changes requested):
Comments on new coding.

roles/baseline/defaults/main.yml
jarpat: From this ticket IAC-445, since this patch was only going to be applied on ingress-nginx 1.1.1 / clusters >= 1.22.x, splitting it out means that deployments on older K8s version would get the original INGRESS_NGINX_CONFIG without these mitigation values.

thpang: Ok, I am ok with that. In the future how will this scale? Right now we have one new section `INGRESS_NGINX_CVE_PATCH` is the plan to add `n` number of these blocks in the future? And then have code in the future that parses these? Trying to figure out this new format.

One other question here is the CVE Patch not applicable to all ingress-nginx versions or is this only for future releases? Was this investigated?

thpang (changes requested):
More questions.

roles/baseline/defaults/main.yml
jarpat: @thpang, It was decided in COMPUTESVCS-27936 (see last comment) that the simplest solution for now was to apply the mitigation based off of the Kubernetes version. This patch will only work on ingress-nginx appVersions v0.50.0 or v1.0.5, or later, see (COMPUTESVCS-28201).

As far as scaling goes, you're right, I can see having multiple patch variables getting messy. In the future when more patches are required I could possibly switch the ingress-nginx configuration generation to be a jinja template. An ansible task would check versions and set Boolean values which would be passed into the template to perform the generation. I could update my implementation now if you think that is better, I'm also open to other suggestions.

thpang: So if our code only works with ingress-nginx 0.5 and 1.0.5 now why have the check? The older code works fine and the code moving forward would have this fix. Not sure the conditional patch etc is needed. And yes it was decided to have the entries but not the solution designed here.

jarpat: @thpang Made the changes after our latest discussion, the default ingress-nginx config was updated rather than having a conditional patching task. See the description where retested these changes in GKE & EKS.

AWSmith0216:
Similar to the prior discussion around the ingress-nginx version to deploy, there's value in making it conditional on K8S version because stable/2021.2.4 is the first cadence that would work with this mitigation in place. However, lts/2021.1, lts/2021.2, stable/2021.2.1, stable/2021.2.2, and stable/2021.2.3 are all currently supported cadences that will not work with this mitigation (per my understanding).

So with this change in place, if someone is baselining a cluster for use with one of those prior cadences, it sounds like they'll need to use a previous release of viya4-deployment.

jarpat:
An older version of viya4-deployment could be used, alternatively a value for INGRESS_NGINX_CONFIG could also be provided

sayeun (approved):

PR #200, opened 2022-03-18
(IAC-478) Dependencies mismatch to viya4-monitoring-kubernetes

riragh:
Updated kubectl version in Dependencies.md from 1.19.9 to a range 1.20 -1.22 to match the viya4-monitoring-kubernetes expected kubectl version.
Test:
- Created Azure cluster with default kubernetes_version: 1.21.7. Viya4 deployment along with logging and monitoring was successful, no issues found in the process.

### DIFF
--- a/docs/user/Dependencies.md
+++ b/docs/user/Dependencies.md
@@ -13,7 +13,7 @@ SOURCE | NAME | VERSION
 ~ | docker | any
 ~ | git | any
 ~ | kustomize | 3.7.0
-~ | kubectl | 1.20+
+~ | kubectl | 1.20 - 1.22
 ~ | AWS IAM Authenticator | 1.18.9/2020-11-02
 ~ | Helm | 3
 pip3 | ansible | 2.10.7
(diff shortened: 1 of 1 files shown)


thpang:
The text info in this issue seems to show that the changes are being made to keep this code in parity with logging and monitoring and that is not the case. This code base is driven by the kubernetes versions that are needed to support SAS Viya 4. Right now those versions are 1.20 - 1.22. Wanted to be sure that was clear.

docs/user/Dependencies.md
```diff
@@ -13,7 +13,7 @@ SOURCE | NAME | VERSION
 ~ | docker | any
 ~ | git | any
 ~ | kustomize | 3.7.0
-~ | kubectl | 1.19.9
+~ | kubectl | 1.20+
```
thpang, line 16: I am kinda ok with this. Right now we support kubernetes 1.20 -1.22 Might want to make this more reflective of that value as folks could run with 1.23 and 1.24 at this time which is not supported; however, the 1.20+ leads one to believe this code is on the leading edge of kubernetes support.

thpang (changes requested):
Check comments

docs/user/Dependencies.md
riragh: Updated the kubectl version to a range 1.20 -1.22.

jarpat (approved):

thpang (approved):
LGTM

PR #202, opened 2022-03-28
(IAC-483) Conditionally Patch ingress-nginx Config

jarpat:
### Changes
The mitigation will be applied by default for K8s clusters whose version is >=1.22.x
and whose ingress-ngnix version is >=1.1.0 or (>=0.50.0 and <1.0.0, for the updated legacy versions of nginx that support this CVE).

### Tests
Performed deployments in AKS on both 1.22.6 & 1.21.9 clusters to see the CVE mitigation patch behavior. All deployments in this table resulted in a healthy Viya deployment.

| Cadence  | K8s Version | Ingress-Nginx Version                      | Was CVE Mitigation Applied by Default |
|----------|-------------|--------------------------------------------|---------------------------------------|
| 2021.2.5 | 1.22.6      | 1.1.0                                      | Yes                                   |
| 2021.2.4 | 1.22.6      | 1.1.0                                      | Yes                                   |
| 2021.2.5 | 1.22.6      | 1.1.1 (Custom INGRESS_NGINX_CHART_VERSION) | Yes                                   |
| Fast R/S | 1.22.6      | 1.1.0                                      | Yes                                   |
| 2021.2.4 | 1.21.9      | 0.50.0                                     | No                                    |
| 2021.2.2 | 1.21.9      | 0.50.0                                     | No                                    |
| 2021.2   | 1.21.9      | 0.50.0                                     | No                                    |

PR #204, opened 2022-04-06
(IAC-333) Adding in hsts-max-age for security compliance.

thpang:


thpang:
Not quite sure why `main` and `staging` are so off such that the merge has conflicts. Will work through these. Will be testing on Wed.

sayeun:
The diff between main and staging is from https://github.com/sassoftware/viya4-deployment/pull/202/files

thpang:
Validated that the config map of the ingress controller now sets the value to approximately 2 years.
```yaml
data:
  allow-snippet-annotations: 'true'
  hsts-max-age: '63072000'
  use-forwarded-headers: 'true'
```

thpang:
Testing now shows with TLS active that the hsts-max-age is now part of the header response:

image

jarpat (approved):
Looks good just need to change the target to staging.

riragh (approved):
LGTM

thpang:
Looking to have this one closed. Not sure what/how the merge branch was changed from staging to main. Guessing since the last release the staging branch was deleted. This might cause issues in the future for work that has been targeted for merge into staging but did not make the cut for that release.

PR #213, opened 2022-04-20
(IAC-525) update ansible task for certificate configuration

jarpat:
## Changes
Update `roles/vdm/tasks/tls.yaml` to correctly apply `cert-manager-provided-ingress-certificate.yaml`. It should only be used if the `V4_CFG_TLS_GENERATOR` is set to cert-manager and the TLS key and Cert are not provided by the user.

Update the customer-provided-merge-sas-certframe-configmap.yaml so that it is correctly applied. It should be used when `V4_CFG_TLS_GENERATOR` is either openssl or cert-manager.

### DIFF
--- a/roles/vdm/tasks/tls.yaml
+++ b/roles/vdm/tasks/tls.yaml
@@ -192,22 +192,35 @@
     - uninstall
     - update
 
-- name: tls - Certificate Generation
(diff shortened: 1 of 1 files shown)


thpang (approved):
LGTM

PR #236, opened 2022-06-08
(IAC-553) move to kubectl 1.22.10 to support k8s 1.23

dhoucgitter:
## Changes
Because Viya will support K8s 1.21, 1.22 and 1.23 in June 2022, the version of kubectl needs to be bumped up to 1.22 so that it will be in the required -1/+1 range for supported versions of K8s.
## Tests

|Cloud Provider|kubectl version|K8s version|Cadence|Deployment Stabilized|
|-------|-------|------|-----|------|
|Azure|1.22.10|1.21.9|Fast:2020|Yes|
|Azure|1.22.10|1.21.7|Fast:2020|Yes|
|Azure|1.22.10|1.22.6|Fast:2020|Yes|
|Azure|1.22.10|1.23.5|Fast:2020|Yes|
|AWS|1.22.10|1.22.9|Fast:2020|InProgress |

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,7 +47,8 @@ ENV HOME=/viya4-deployment
 RUN pip install -r ./requirements.txt \
   && ansible-galaxy install -r ./requirements.yaml \
   && chmod -R g=u /etc/passwd /etc/group /viya4-deployment/ \
-  && chmod 755 /viya4-deployment/docker-entrypoint.sh
+  && chmod 755 /viya4-deployment/docker-entrypoint.sh \
+  && git config --system --add safe.directory /viya4-deployment
 
 ENV PLAYBOOK=playbook.yaml
 ENV VIYA4_DEPLOYMENT_TOOLING=docker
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -45,8 +47,7 @@ ENV HOME=/viya4-deployment
 RUN pip install -r ./requirements.txt \
   && ansible-galaxy install -r ./requirements.yaml \
   && chmod -R g=u /etc/passwd /etc/group /viya4-deployment/ \
-  && chmod 755 /viya4-deployment/docker-entrypoint.sh \
-  && git config --system --add safe.directory /viya4-deployment
```
thpang, line 49: Why was this removed? If someone clones this repo it'll have a value.

thpang (commented):
Looking for an answer on question

Dockerfile
dhoucgitter: I removed line 49 since for this Dockerfile the /viya4-deployment folder is not a git managed folder and when I ran "docker build" with that line in place, I got a git error indicating that. Removing the line allowed me to build the container. Could it be an issue with the version of git I'm using?

dhoucgitter: @thpang, I restored the deleted line after our discusion, ready for re-review

thpang (approved):
LGTM

jarpat (approved):
👍

PR #245, opened 2022-06-21
feat: (IAC-480) Change default certificate generator to openssl

jarpat:
# Changes
Makes the new default for `V4_CFG_TLS_GENERATOR` to openssl. Also changes `CERT_MANAGER_ENABLED` to false so cert-manager does not get installed during the baseline by default.

# Tests
See internal IAC-480 ticket

### DIFF
--- a/docs/CONFIG-VARS.md
+++ b/docs/CONFIG-VARS.md
@@ -183,7 +183,7 @@ Viya 4 supports 2 different types of certificate generators, Cert-manager and op
 
 | Name | Description | Type | Default | Required | Notes | Tasks |
 | :--- | ---: | ---: | ---: | ---: | ---: | ---: |
-| V4_CFG_TLS_GENERATOR | Which tool to use for certificate generation | string | cert-manager | false | Supported values: [`cert-manager`,`openssl`]. | viya, cluster-logging, cluster-monitoring |
+| V4_CFG_TLS_GENERATOR | Which tool to use for certificate generation | string | openssl | false | Supported values: [`cert-manager`,`openssl`]. | viya, cluster-logging, cluster-monitoring |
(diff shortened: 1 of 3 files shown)


thpang (commented):
Just a general comment here. There is no parity between what used to be in place for the cert-manager and openssl that was installed during the deployment repo install phase. Would be nice, if possible to have openssl as part of the baseline to keep things aligned.

jarpat:
> Just a general comment here. There is no parity between what used to be in place for the cert-manager and openssl that was installed during the deployment repo install phase. Would be nice, if possible to have openssl as part of the baseline to keep things aligned.

I think this is ready for re-review after the discussion we had last week.

docs/CONFIG-VARS.md
```diff
@@ -279,7 +279,7 @@ V4_CFG_POSTGRES_SERVERS:
 
 | Name | Description | Type | Default | Required | Notes | Tasks |
 | :--- | ---: | ---: | ---: | ---: | ---: | ---: |
-| CERT_MANAGER_ENABLED | Whether to deploy tool | bool | true | false | | baseline |
+| CERT_MANAGER_ENABLED | Whether to deploy tool | bool | false | false | | baseline |
```
dhoucgitter, line 282: Just a doc comment.
Since there are now two certificate handling objects that IAC can impact, I think it might help to say, "Whether to deploy certificate-manager into the cluster" to be more specific than "tool" here.

jarpat: @dhoucgitter, you're right that description doesn't seem to be clear. I've updated the description and notes section to:

| CERT_MANAGER_ENABLED | Whether to deploy cert-manager into the cluster using helm | bool | false | false | Required if V4_CFG_TLS_GENERATOR is set to `cert-manager` and it's not already installed | baseline |

Let me know if that sounds better

PR #258, opened 2022-07-17
fix: (IAC-480) Add Default TLS Generator Variable for Monitoring

jarpat:
## Changes
The monitoring role did not include `V4_CFG_TLS_GENERATOR` in its default variables file. This causes the cluster-monitoring deployment task to fail if the user tries to use the upcoming viya4-monioring release (master branch) when they do not include that variable in their ansible-vars.yaml.

## Tests
After the changes were put in, I performed a deployment with the following tags "baseline,viya,cluster-logging,cluster-monitoring,viya-monitoring,install". I ensured that `V4_CFG_TLS_GENERATOR` was removed from my ansible-vars.yaml file to verify that the default variable gets used.

See internal ticket for test artifacts.


roles/monitoring/defaults/main.yaml
```diff
@@ -37,3 +37,5 @@ V4M_GRAFANA_FQDN: "grafana.{{ V4M_BASE_DOMAIN }}"
 V4M_GRAFANA_CERT: "{{ V4M_CERT }}"
 V4M_GRAFANA_KEY: "{{ V4M_KEY }}"
 V4M_GRAFANA_PASSWORD: "{{ lookup('password', '/dev/null chars=ascii_letters,digits') }}"
+
+V4_CFG_TLS_GENERATOR: openssl # [cert-manager,openssl]
```
thpang, line 41: Why is this here? It's a generic V4 variable and should already be defined with a value. This should not be added here. You'll have to verify that the variable you're looking for is defined properly and has a default. This is not the place.

thpang (changes requested):
Need to rework logic

riragh (approved):
LGTM

PR #265, opened 2022-07-27
fix: (IAC-556) Incorrect tls.key in alertmanager-ingress-tls-secret

dhoucgitter:
Issue Description:
Reported by Adam Bullock, "a recent deployment with monitoring and logging enabled, that the tls.key in the alertmanager-ingress-tls-secret was incorrect. The certificate was added instead of the key."

The following messages are logged in the ingress-nginx-controller log:

Error obtaining X.509 certificate: no object matching key "monitoring/alertmanager-ingress-tls-secret" in local store Error getting SSL certificate "monitoring/alertmanager-ingress-tls-secret": local SSL certificate monitoring/alertmanager-ingress-tls-secret was not found. Using default certificate

The problematic line: https://github.com/sassoftware/viya4-deployment/blob/main/roles/monitoring/tasks/cluster-monitoring.yaml#L118

Rationale for setting the the delete namespace timeouts to 10 mins is that I found existing namespace operations that were already set to use a 10 minute timeout so it made sense to keep them all the same.

### DIFF
--- a/roles/monitoring/tasks/cluster-logging.yaml
+++ b/roles/monitoring/tasks/cluster-logging.yaml
@@ -140,6 +140,7 @@
     kind: Namespace
     name: "{{ V4M_LOGGING_NAMESPACE }}"
     wait: true
+    wait_timeout: 600
     kubeconfig: "{{ KUBECONFIG }}"
     state: absent
   tags:
--- a/roles/monitoring/tasks/cluster-monitoring.yaml
+++ b/roles/monitoring/tasks/cluster-monitoring.yaml
@@ -121,7 +121,7 @@
         tls.crt: >-
           {{ lookup('file', V4M_ALERTMANAGER_CERT ) |b64encode }}
         tls.key: >-
-          {{ lookup('file', V4M_ALERTMANAGER_CERT ) |b64encode }}
+          {{ lookup('file', V4M_ALERTMANAGER_KEY ) |b64encode }}
       type: kubernetes.io/tls
   when: 
     - V4M_ALERTMANAGER_CERT is not none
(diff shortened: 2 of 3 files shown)


riragh (approved):
LGTM

roles/monitoring/tasks/cluster-logging.yaml
```diff
@@ -140,6 +140,7 @@
     kind: Namespace
     name: "{{ V4M_LOGGING_NAMESPACE }}"
     wait: true
+    wait_timeout: 600
```
thpang, line 143: Is this something we can create a variable for or have already? Also is this in seconds? Needs to be clear if we're exposing this to the customer. If not this is fine.

roles/monitoring/tasks/cluster-monitoring.yaml
```diff
@@ -143,6 +143,7 @@
     kind: Namespace
     name: "{{ V4M_MONITORING_NAMESPACE }}"
     wait: true
+    wait_timeout: 600
```
thpang, line 146: Same comment as above.

thpang (commented):
Some general comments. Not sure there's a code change here.

roles/monitoring/tasks/cluster-logging.yaml
dhoucgitter: We don't have a variable for this already but wondered that also and it is seconds. 
Three other namespace delete task timeouts just like these already existed in other sections of the code before the one update and two new ones I added in this PR for monitoring and logging. I searched the current code and docs and see no mention of these timeouts so would conclude that we don't expose it.

thpang: That's what I needed to know. Looks good then.

thpang (approved):
LGTM

PR #272, opened 2022-08-04
feat: (IAC-564) SAS Viya Startup Sequencer - Add entry in kustomization.yaml

riragh:
# Changes:
Enabled the startup sequencer feature by referencing the overlay "sas-bases/overlays/startup/ordered-startup-transformer.yaml " in the transformers section of kustomization.yaml before the line that includes "sas-bases/overlays/required/transformers.yaml".

# Test:
See internal ticket for details. Performed two deployments one with small VA order and another large NA order. Both deployments had stabilization times improved.


thpang (approved):
LGTM

dhoucgitter (approved):
lgtm

PR #277, opened 2022-08-09
feat: (IAC-629) Support SingleStore (DAC)

dhoucgitter:
Add SingleStore README and siteconfig_info change
- Add a link to SingleStore requirements in SAS Viya Operations Guide 2022.1.4
- Add a link to SingleStore README from base README.md to expose it to DAC consumers
- Add additional step and step to skip for sas-bases/examples/sas-singlestore/README.md
- update siteconfig_info to use folder name instead of full path to kustomization.yaml file for component entries

### DIFF
--- a/docs/user/SingleStore.md
+++ b/docs/user/SingleStore.md
@@ -1,0 +1,27 @@
+# SAS Viya Support for SingleStore
+
+SAS Viya supports SingleStore. SingleStore is a cloud-native database designed for data-intensive applications. A distributed, relational, SQL database management system that features ANSI SQL support, it is known for speed in data ingest, transaction processing, and query processing. 
+
+## Requirements for a SingleStore SAS Viya deployment
+
+If your SAS software order included SAS Viya with SingleStore, there are additional requirements that may apply to your deployment. The SAS Viya _IT Operations Guide_ provides detailed information about requirements for a SingleStore enabled deployment of SAS Viya. Access it [here](https://documentation.sas.com/doc/en/itopscdc/default/itopssr/n0jq6u1duu7sqnn13cwzecyt475u.htm#n0qs42c42o8jjzn12ib4276fk7pb)
+
+## Deploying SingleStore with SAS Viya Deployment
+
+You can deploy SAS Viya with SingleStore into a Kubernetes cluster that is running in Microsoft Azure. The [SAS Viya 4 Infrastructure as Code (IaC) for Microsoft Azure](https://github.com/sassoftware/viya4-iac-azure) GitHub project can automatically provision the required infrastructure components that support SAS Viya with SingleStore deployments. Refer to the [SingleStore sample input file](https://github.com/sassoftware/viya4-iac-azure/blob/singlestore/examples/sample-input-singlestore.tfvars) for terraform configuration values that create an Azure cluster suitable for deploying Viya and SingleStore to.
+
+## Customizing SingleStore Deployment Overlays
+
+Refer to the SAS Viya Deployment instructions [Getting Started](https://github.com/sassoftware/viya4-deployment#getting-started) and the [SAS Viya Customizations](https://github.com/sassoftware/viya4-deployment#sas-viya-customizations) section for information on how users can make changes to their deployment by adding custom overlays into subfolders under the `/site-config` folder.
+
+After running SAS Viya Deployment with the setting DEPLOY=false in your ansible-vars.yaml file, locate the `sas-bases/` folder which is a peer to the `site-config/` folder underneath your deployments <base_dir>.
+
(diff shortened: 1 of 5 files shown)


roles/vdm/library/siteconfig_info.py
```diff
@@ -62,7 +62,7 @@ def traverse(self, folder):
           yamlblock = yaml.safe_load(file)
           # handle components
           if "kind" in yamlblock and yamlblock['kind'] == "Component":
-            self.add_overlays(Overlay.COMPONENT, kustomizefilefullpath)
+            self.add_overlays(Overlay.COMPONENT, folder)
```
thpang, line 65: The code now forces all component level items to be in their own folder with this code update correct? Not saying this is a bad thing, but we have to make sure any older code that might have included a component that lives in say for example: site-config/foo/mycomponent.yaml file is also still processed correctly if this is a viable example.

This would be a major change as we're now altering how DAC works within the structure of the site-config directory.

thpang (commented):
Need feedback on changes to the code processing for components.

roles/vdm/library/siteconfig_info.py
dhoucgitter: It does require that component level items be in their own folder which I do not believe is a bad thing. Checking the examples of components that Viya provides, there are 18 of those under (security, sas-cas-server and sas-singlestore, in my order) all follow the pattern of providing a component within a folder and that folder contains one kustomization.yaml file.
Based on experience of what a reference should look like for a component or resource in a base kustomization.yaml file, the base kustomization.yaml entries for components that were created before the change should generate a kustomize error (generating a manifest) if someone had tried to include a different component by copying it into the site-config folder structure. I'll run a test to verify that conclusion.

thpang: If the previous items were setup under their own component directories was this working or do we have hardcoded items in DAC picking these items up or adding the entries into the kustomization.yaml file? Not sure we were actually processing these directories. Let me know what you find

dhoucgitter: yes, we have hardcoded items in DAC placing the correct entries for components in the base kustomization.yaml file, TLS is one example, so those are not using the code in siteconfig_info that was generating the improper entries.

dhoucgitter: I posted some test results in our team chat.

roles/vdm/templates/kustomization.yaml
```diff
@@ -25,4 +25,12 @@ namespace: {{ NAMESPACE }}
 {% endif %}
 
 {% endfor %}
-
+{% if "components" not in ordered_overlays.result %}
```
dhoucgitter, line 28: This update to the kustomization.yaml template handles the scenario when DAC does not contribute any hard-coded component entries into ordered_overlays (as when TLS is disabled) and then skips over adding user submitted components under site-config such as SingleStore.

thpang (approved):
LGTM

docs/user/SingleStore.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Requirements for a SingleStore SAS Viya deployment
+
+If your SAS software order included SAS Viya with SingleStore, there are additional requirements that may apply to your deployment. The SAS Viya _IT Operations Guide_ provides detailed information about requirements for a SingleStore enabled deployment of SAS Viya. Access it [here](https://documentation.sas.com/doc/en/itopscdc/default/itopssr/n0jq6u1duu7sqnn13cwzecyt475u.htm#n0qs42c42o8jjzn12ib4276fk7pb)
+
+## Deploying SingleStore with SAS Viya Deployment
+
+You can deploy SAS Viya with SingleStore into a Kubernetes cluster that is running in Microsoft Azure. The [SAS Viya 4 Infrastructure as Code (IaC) for Microsoft Azure](https://github.com/sassoftware/viya4-iac-azure) GitHub project can automatically provision the required infrastructure components that support SAS Viya with SingleStore deployments. Refer to the [SingleStore sample input file](https://github.com/sassoftware/viya4-iac-azure/blob/singlestore/examples/sample-input-singlestore.tfvars) for terraform configuration values that create an Azure cluster suitable for deploying Viya and SingleStore to.
```
jarpat, line 11: Change 'singlestone' in link to 'main' so it points to correct file once merged.
```suggestion
You can deploy SAS Viya with SingleStore into a Kubernetes cluster that is running in Microsoft Azure. The [SAS Viya 4 Infrastructure as Code (IaC) for Microsoft Azure](https://github.com/sassoftware/viya4-iac-azure) GitHub project can automatically provision the required infrastructure components that support SAS Viya with SingleStore deployments. Refer to the [SingleStore sample input file](https://github.com/sassoftware/viya4-iac-azure/blob/main/examples/sample-input-singlestore.tfvars) for terraform configuration values that create an Azure cluster suitable for deploying Viya and SingleStore to.
```

jarpat (approved):
👍

PR #278, opened 2022-08-09
feat: (IAC-658): Support CAS customizations for tenants

riragh:
# Changes:
- Updated the CAS transformer templates on provider side to be restricted to default CAS server
- Added a new map of object variable V4MT_TENANT_CAS_CUSTOMIZATION to accept CAS customization for the tenants
See the new structure in CONFIG-VARS.md file

# Tests
- Verified provider deployment successful with updated CAS transformers.
- Verified different combinations of the tenant CAS customizations by setting V4MT_TENANT_CAS_CUSTOMIZATION.
- Verified tenant CAS server directories are created with default setting if customizations are not provided.


docs/CONFIG-VARS.md
```diff
@@ -345,5 +345,45 @@ The nfs-client is currently supported by the newer nfs-subdir-external-provision
...
+
+```bash
+V4MT_TENANT_CAS_CUSTOMIZATION:
+  ACME:
+    V4MT_CFG_CAS_RAM: 3Gi
+    V4MT_CFG_CAS_CORES: 300m
+    V4MT_CFG_CAS_ENABLE_LOADBALANCER: true
+    V4MT_LOADBALANCER_SOURCE_RANGES: ['0.0.0.0/0']
+    V4MT_CFG_CAS_WORKER_COUNT: 0
+    V4MT_CFG_CAS_ENABLE_BACKUP_CONTROLLER: false
```
thpang, line 383: These are not V4MT varaibles at this point the customer is simply applying those so they should look like the following:

```bash
V4MT_TENANT_CAS_CUSTOMIZATION:
  acme:
    cpus: 4
    memory: 8
    loadbalancer_enabled: true
    loadbalancer_source_ranges: ['0.0.0.0/0']
    backup_controller_enabled: false
    worker_count: 0
```
Then in the code we map things over to whatever we need internally. This is much cleaner for the end user to understand if that's possible given the complexity of these variables. You'll need to update the descriptions to let folks know if they are assigning actual CPUs, memory, etc or if those are system resource requests from kubernetes which is very different.

Also wondering about the load balancer, is that setup by namespace or do all these tenants live in the same namespace where traffic can cross?

roles/multi-tenancy/defaults/main.yml
```diff
@@ -13,13 +13,13 @@ V4MT_MODE: "schema"
 
 V4MT_TENANT_IDS: null
 
-# Optional job paramter
+# Optional job parameter
 V4MT_PROVIDER_PASSWORD: null
 
-V4MT_CAS_WORKER_COUNT: 0
-V4MT_CAS_BACKUP: 0
```
thpang, line 20: Are these breaking anything be removing them? I know we'll be removing the provider based CAS soon, but having these values removed now seems a bit premature. Should they not stay? I don't see equivalent items in the code. Until the base CAS provider has been removed all items related to that code base that are relevant must stay.

thpang (changes requested):
Initial comments and changes

docs/CONFIG-VARS.md
riragh: I'll update the variables to match the structure you mentioned.
All the tenants live in the same namespace as Viya. From the CAS documentation looks like you could use the cas-enable-external-services.yaml to set different loadbalancer source ranges for different CAS servers. In the current setup I could only verify that the values that are specified got assigned correctly. Not sure, maybe testing team could do intensive test around this.

roles/multi-tenancy/defaults/main.yml
riragh: No nothing is breaking by removing this variables. These were moved over to be part of the new variable. Under V4MT_TENANT_CAS_CUSTOMIZATION see: V4MT_CFG_CAS_WORKER_COUNT, V4MT_CFG_CAS_ENABLE_BACKUP_CONTROLLER

docs/CONFIG-VARS.md
riragh: About description, these variables under the exactly same as on the provider side with same description. The only difference is they are for tenants. See the original provider CAS variables here:
https://github.com/sassoftware/viya4-deployment/blob/staging/docs/CONFIG-VARS.md#cas
Let me if we still need any further updates on descriptions.

roles/multi-tenancy/defaults/main.yml
thpang: So the provider based CAS server does not use or have worker and or backups? Again, asking as this was there for a reason. So guessing that the provider does need these until CAS does not have this in play during multi-tenancy correct? Just covering the bases as folks might not employ multi-tenancy so these would be valid.

docs/CONFIG-VARS.md
thpang: I agree with the description. Those are fine. We've already been telling folks the same info.

roles/multi-tenancy/defaults/main.yml
riragh: These are not the provider variables. The provider side variables V4_CFG_CAS_WORKER_COUNT and V4_CFG_CAS_ENABLE_BACKUP_CONTROLLER are still in place and should not be affected. The variables I removed and moved were tenant specific starting with 'V4MT'.

thpang: Got it. Sorry should have looked at the prefix more closely. You did identify these as V4MT items. Thx.

docs/CONFIG-VARS.md
```diff
@@ -345,5 +345,45 @@ The nfs-client is currently supported by the newer nfs-subdir-external-provision
...
\ No newline at end of file
+| V4MT_TENANT_CAS_CUSTOMIZATION | Map of objects with all tenant CAS customization variables. See the format below | | | false | | multi-tenancy |
+
+### Tenant CAS Customization
+
+Some of the tenant CAS customizations can be defined with the V4MT_TENANT_CAS_CUSTOMIZATION variable which is a map of objects. The variable has the following format:
+
+```bash
+V4MT_TENANT_CAS_CUSTOMIZATION:
+  <TENANT-ID1>:
```
thpang, line 356: Would in general use lower case items as part of the example here. Nothing major

roles/multi-tenancy/tasks/tenant-cas-customizations.yaml
```diff
@@ -0,0 +1,79 @@
+---
+- name: update tenant cas customizations
+  block:
+    - name: tenant cas - MPP workers
```
thpang, line 4: Are these truly MPP workers? Just want to verify or if that's the only way to have multi-tenancy support.

roles/multi-tenancy/templates/cas-manage-cpu-and-memory.yaml
```diff
@@ -0,0 +1,46 @@
...
+    value:
+      memory: {{ AMOUNT-OF-RAM }}
+  - op: replace
+    path: /spec/controllerTemplate/spec/containers/0/resources/requests/memory
+    value:
+      {{ AMOUNT-OF-RAM }}
+  - op: add
+    path: /spec/controllerTemplate/spec/containers/0/resources/limits/cpu
+    value:
+      {{ NUMBER-OF-CORES }}
```
thpang, line 32: Are the limits really the same as the requested number provided? Just checking. Most deployments have a set value for the upper limit that's independent of the requested limit.

roles/multi-tenancy/templates/cas-manage-cpu-and-memory.yaml
```diff
@@ -0,0 +1,46 @@
...
+---
+apiVersion: builtin
+kind: PatchTransformer
+metadata:
+  name: cas-manage-cpu-and-memory
+patch: |-
+  - op: add
+    path: /spec/controllerTemplate/spec/containers/0/resources/limits
+    value:
+      memory: {{ AMOUNT-OF-RAM }}
```
thpang, line 24: Are the limits really the same as the requested number provided? Just checking. Most deployments have a set value for the upper limit that's independent of the requested limit.

roles/multi-tenancy/templates/cas-manage-cpu-and-memory.yaml
```diff
@@ -0,0 +1,46 @@
...
+  - op: replace
+    path: /spec/controllerTemplate/spec/containers/0/resources/requests/cpu
+    value:
+      {{ NUMBER-OF-CORES }}
+target:
+  group: viya.sas.com
+  kind: CASDeployment
+  # Uncomment this to apply to all CAS servers:
+  # name: .*
+  # This applies to one particular named CAS server:
```
thpang, line 42: Lots of stray comments here. Those will be removed before production correct?

roles/multi-tenancy/tasks/tenant-cas-customizations.yaml
```diff
@@ -0,0 +1,79 @@
...
+        cmd: "{{ DEPLOY_DIR }}/site-config/create-cas-server.sh --tenant {{ tenant | lower }} \
+             --output {{ DEPLOY_DIR }}/site-config \
+             --workers {{ settings.worker_count if (settings.worker_count is defined and settings.worker_count is not none ) else 0 }} \
+             --backup {{ 1 if (settings.backup_controller_enabled is defined) and \
+                              (settings.backup_controller_enabled|bool == True) else 0 }}"
+
+    - name: tenant cas user-defined resources - copy template
+      copy:
+        src: "{{ role_path }}/templates/cas-manage-cpu-and-memory.yaml"
+        dest: "{{ DEPLOY_DIR }}/site-config/cas-{{ tenant | lower }}-default/cas-manage-cpu-and-memory.yaml"
```
thpang, line 15: In the future it may be easier to set a fact that contains the tenant or variable in its final state so here you could have `_tenant` set to `{{ tenant | lower }}` which would only require the conversion once.

thpang (changes requested):
Only a couple of updates requested. Also what's the focus on the store for each tenant? Is there any guidance setup that the CAS documentation clarifies? Not seeing any information here on that.

roles/multi-tenancy/tasks/tenant-cas-customizations.yaml
riragh: This task is to set the tenant CAS to MPP when the worker_count is specified. Same as provider, it's the MPP support for tenants.

roles/multi-tenancy/templates/cas-manage-cpu-and-memory.yaml
riragh: It didn't find much guidance which clarifies the information.
Also, looking at the transformer file and later in the pod description, the number provided is set as the maximum requested limit.

roles/multi-tenancy/templates/cas-manage-cpu-and-memory.yaml
riragh: same response as above.

roles/multi-tenancy/tasks/tenant-cas-customizations.yaml
riragh: Updated in the latest change. Instead of setting a fact, I added conversion when key is passed to the task. So now lower conversion only occurs at one place.

roles/multi-tenancy/tasks/tenant-cas-customizations.yaml
thpang: Ok, where's the logic with the check for `worker_count`? Not seeing it.

roles/multi-tenancy/templates/cas-manage-cpu-and-memory.yaml
thpang: If there is no guidance it's a good best guess this point. I am fine with this.

roles/multi-tenancy/tasks/tenant-cas-customizations.yaml
riragh: See line number 8.
From the branch: https://github.com/sassoftware/viya4-deployment/blob/cas_updates/roles/multi-tenancy/tasks/tenant-cas-customizations.yaml#L8

thpang: thx.

thpang (approved):
LGTM. Thanks for the updates

docs/CONFIG-VARS.md
riragh: Done.

dhoucgitter (commented):
lgtm, content in tenant-cas-customizations.yaml was interesting

dhoucgitter (approved):
meant to click on approve b4

PR #279, opened 2022-08-10
Added ability to append custom definitions to kustomization.yaml.

### LINKED ISSUE #223: Customization for relocating SASWORK is not treated properly
This is further documented in tech support track: https://sirius.na.sas.com/Sirius/GSTS/ShowTrack.aspx?trknum=7613571759

I am trying to ensure that SASWORK is pointing to a hostPath rather than the emptyDir that it points to by default.  The approach to this is supposed to be modification of the change-viya-volume-storage-class.yaml file located at: sas-bases/examples/sas-programming-environment/storage.  FYI, I'm working from the documentation here: https://go.documentation.sas.com/doc/en/sasadmincdc/v_024/calsrvpgm/n0k315nlna2awln119phkhua1na2.htm on a 2021.2.4 deployment.

When I copy that file into  a directory under site-config and modify it for my hostPath, site.yaml will not build.  It appears that the VIya 4 Deployment project picks up the customization, but doesn't represent it properly in kustomization.yaml (it's treating a patch as a resource according to tech support).

What appears in kustomization.yaml is:

resources:
...
  - site-config/sas-programming-environment/change-viya-volume-storage-class.yaml

But what it *should* be putting in kustomization.yaml (again, according to tech support) is:

patches:
  - path: site-config/change-viya-volume-storage-class.yaml
    target:
      kind: PodTemplate
      labelSelector: "sas.com/template-intent=sas-launcher"

I have been able to work around this (I think), as I have gotten site.yaml to build but I can't attempt applying it until tonight when my environment's users are finished for the day.  I'm working around by doing the following:

1. I moved the change-viya-volume-storage-class.yaml file out of the site-config directory so that it does not get picked up by the ansible job.
2. I've changed my ansible-vars file so that deploy is set to false.
3. I've created a bash script that run the ansible playbook, then appends the "patches" block from above to the end of the kustomization.yaml file that ansible has created, and finally rebuilds site.yaml based on this modified kustomization.yaml file.

While I'm optimistic that this approach will work, every customer deployment should be leveraging this patch as the emptyDir default can cause all sorts of problems and it seems like there should be an easier way.   Relocating SASWORK should not require re-engineering of the whole deployment process.  Is there a different approach I should be taking entirely?

sweosm:
To solve the issue with SASWORK location and similar customizations, I made some changes to allow for a custom block to be added to the end of kustomization.yaml.

This fixes #223

Instructions:

1. Add a folder "customkustomize" to `<cluster>/<namespace>/site-config/`
2. Add the files you want to put in the kustomization.yaml file that's not under "resources" to this new folder
3. Add a file "customkustomize.yaml" to that folder which contains exactly what you want to append in the resulting kustomization.yaml file.

The end result can for example look like this:
```
$ ls site-config/customkustomize/
change-viya-volume-storage-class.yaml  customkustomize.yaml
$ cat site-config/customkustomize/customkustomize.yaml
patches:
  - path: site-config/customkustomize/change-viya-volume-storage-class.yaml
    target:
      kind: PodTemplate
      labelSelector: "sas.com/template-intent=sas-launcher"
```

### DIFF
--- a/roles/vdm/library/siteconfig_info.py
+++ b/roles/vdm/library/siteconfig_info.py
@@ -96,6 +96,12 @@ class siteConfig(object):
     for yamlfile in yamlfiles:
       self.addResource(yamlfile)
 
+  def add_customkustomize_block(self, folder):
+    if os.path.exists(os.path.join(folder, "customkustomize.yaml")):
+      kustomizeCustomfilefullpath = os.path.join(folder, "customkustomize.yaml")
+      with open(kustomizeCustomfilefullpath) as file:
+        self._overlays['customkustomize']  = file.read()
+
 def main():
   fields = {
     "path": {"required": True, "type": "str"},
@@ -111,6 +117,9 @@ def main():
       for exclude in module.params['exclude']:
         if folder == exclude:
           skip = True
+      if folder == "customkustomize":
+        sc.add_customkustomize_block(os.path.join(scFolder, folder))
+        skip = True
       if not skip:
         sc.traverse(os.path.join(scFolder, folder))
     module.exit_json(changed=True, overlays=sc.get_overlays())
--- a/roles/vdm/templates/kustomization.yaml
+++ b/roles/vdm/templates/kustomization.yaml
@@ -26,3 +26,6 @@ namespace: {{ NAMESPACE }}
 
 {% endfor %}
 
+{% if 'customkustomize' in user_customizations.overlays %}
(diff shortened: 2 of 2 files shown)


riragh:
Closed after deleting staging, reopening again

thpang:
Closing this one as code base has changed and the current work-flow allows the customer to add items in the site-config directory with customer code.

PR #281, opened 2022-08-11
(IAC-658): fix: Updating default value for V4MT_TENANT_CAS_CUSTOMIZATION

riragh:
Originally the default value for V4MT_TENANT_CAS_CUSTOMIZATION was set as null. However, ansible treats 'null' as a valid value for dictionary being defined and tries to run dictionary operations.

# Change: 
Updated the default value for V4MT_TENANT_CAS_CUSTOMIZATION as a empty dictionary.

# Tests:
This specific issue was hit when V4MT_TENANT_CAS_CUSTOMIZATION wasn't provided and ansible used the null value as default.
- Verified with the default changed, the issue was not seen. 
- Ran other combinations with V4MT_TENANT_CAS_CUSTOMIZATION only defined for two out three tenants
- Also verified when V4MT_TENANT_CAS_CUSTOMIZATION  is defined with only tenant names but no customizations.

### DIFF
--- a/roles/multi-tenancy/defaults/main.yml
+++ b/roles/multi-tenancy/defaults/main.yml
@@ -17,7 +17,7 @@ V4MT_TENANT_IDS: null
 V4MT_PROVIDER_PASSWORD: null
 
 # CAS customization values
-V4MT_TENANT_CAS_CUSTOMIZATION: null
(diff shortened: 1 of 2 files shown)


dhoucgitter (approved):
lgtm

thpang (approved):
LGTM. Thanks for the clarification

PR #285, opened 2022-08-23
fix: (IAC-670) Correct case typo for alertmanagerSpec

dhoucgitter:
A yaml heading (alertManagerSpec) for alert manager didn't match the resource definition heading for alert manager in the helm chart used to deploy it. That mismatch resulted in the V4M_STORAGECLASS configuration variable setting being ignored for alert manager.


thpang (approved):
LGTM

riragh (approved):
LGTM

PR #287, opened 2022-08-26
fix: (IAC-689) add new exclusion for storageclass patch transformer

dhoucgitter:
Why:

Release testready orders now include the du sas-data-agent-server-colocated
which does not have /spec/VolumeClaimTemplates in its StatefulSet
resulting in fatal kustomize error during DAC execution. Excluding that du
from the patchTransformer as we've done for two other dus earlier.


jarpat (approved):

sayeun (approved):
Looks good.

riragh (approved):
Looks good

thpang (approved):
Yup, that's the pattern. LGTM

PR #293, opened 2022-09-07
feat: (IAC-582) Support configuring schedule-start-stop.yaml

jarpat:
### Changes

Allow the user to configure and include the schedule-start-stop.yaml transformer using the two new variables `V4_CFG_VIYA_STOP_SCHEDULE` & `V4_CFG_VIYA_START_SCHEDULE`. This configures cronjobs in Kubernetes that will start and stop your Viya deployment using the times you specified for these variables. As per the deployment instructions this transformer is prioritized to be placed after `sas-bases/overlays/required/transformers.yaml` in the kustomization.yaml

### Tests

See internal ticket for more details/artifacts

1. In a 1.23.8 AKS cluster performed a Viya deployment on the latest cadence with `V4_CFG_VIYA_STOP_SCHEDULE: "0 19 * * 1-5"` & `V4_CFG_VIYA_START_SCHEDULE: "30 19 * * 1-5"` set. Verified the positioning of the transformer in the kustomization.yaml. Also verified the Viya deployment successfully stopped at 19:00 UTC and started back up at 19:30 UTC.
2. Input validation scenarios, ensured that `schedule-start-stop.yaml` was only used if using a supported cadence and if both variables are defined.

### DIFF
--- a/docs/CONFIG-VARS.md
+++ b/docs/CONFIG-VARS.md
@@ -274,6 +274,8 @@ V4_CFG_POSTGRES_SERVERS:
 | V4_CFG_EMBEDDED_LDAP_ENABLE | Deploy openldap in the namespace for authentication | bool | false | false | [Openldap Config](../roles/vdm/templates/generators/openldap-bootstrap-config.yaml) | viya |
 | V4_CFG_CONSUL_ENABLE_LOADBALANCER | Setup LB to access consul ui | bool | false | false | Consul ui port is 8500 | viya |
 | V4_CFG_ELASTICSEARCH_ENABLE | Enable opendistro search | bool | true | false | When deploying LTS less than 2020.1 or Stable less than 2020.1.2 set to false | viya |
+| V4_CFG_VIYA_START_SCHEDULE | Configure your SAS Viya deployment to start on specific schedules | string |  | false | Must be used in conjunction with `V4_CFG_VIYA_STOP_SCHEDULE` | viya |
+| V4_CFG_VIYA_STOP_SCHEDULE | Configure your SAS Viya deployment to stop on specific schedules | string |  | false | Must be used in conjunction with `V4_CFG_VIYA_START_SCHEDULE` | viya |
 
 ## 3rd Party tools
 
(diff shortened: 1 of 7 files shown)


roles/vdm/tasks/start_stop.yaml
```diff
@@ -0,0 +1,42 @@
+- name: start_stop - validate cadence version
+  ansible.builtin.fail:
+    msg: >
+      The schedule-start-stop.yaml transformer is not supported is not supported for cadences 2021.1 and older
```
riragh, line 4: minor change request: Message needs update `is not supported` was repeated.

docs/CONFIG-VARS.md
```diff
@@ -274,6 +274,8 @@ V4_CFG_POSTGRES_SERVERS:
...
 | V4_CFG_CONSUL_ENABLE_LOADBALANCER | Setup LB to access consul ui | bool | false | false | Consul ui port is 8500 | viya |
 | V4_CFG_ELASTICSEARCH_ENABLE | Enable opendistro search | bool | true | false | When deploying LTS less than 2020.1 or Stable less than 2020.1.2 set to false | viya |
+| V4_CFG_VIYA_START_SCHEDULE | Configure your SAS Viya deployment to start on specific schedules | string |  | false | This variable accepts [CronJob schedule expressions](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-schedule-syntax) to create your Viya start job schedule. Must be used in conjunction with `V4_CFG_VIYA_STOP_SCHEDULE` | viya |
+| V4_CFG_VIYA_STOP_SCHEDULE | Configure your SAS Viya deployment to stop on specific schedules | string |  | false | This variable accepts [CronJob schedule expressions](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-schedule-syntax) to create your Viya stop job schedule. Must be used in conjunction with `V4_CFG_VIYA_START_SCHEDULE` | viya |
```
thpang, line 278: Are there samples of this for the end user? I see the links, but having an example file with these values would be helpful.

thpang (changes requested):

roles/vdm/tasks/start_stop.yaml
jarpat: Good catch, fixed.

docs/CONFIG-VARS.md
jarpat: I can add something in the example files for these two variables. I'll go a with a similar route to "Monitoring and Logging" in the existing example files:
https://github.com/sassoftware/viya4-deployment/blob/main/examples/ansible-vars-iac.yaml#L36

Include example variables but comment them out, since the user would not necessarily want to start and stop Viya on schedule by default.

jarpat: Updated example files, ready for re-review.

riragh (approved):
LGTM

dhoucgitter (approved):
lgtm

docs/CONFIG-VARS.md
thpang: Will take a look. Thx

thpang (approved):
LGTM

PR #294, opened 2022-09-12
feat: (IAC-582) Allow user to specify either V4_CFG_VIYA_START_SCHEDULE or V4_CFG_VIYA_STOP_SCHEDULE

jarpat:
### Changes
Allow the user to specify `V4_CFG_VIYA_START_SCHEDULE` or `V4_CFG_VIYA_STOP_SCHEDULE` rather than requiring both when using the start-stop transformer.
For the variable that the user chooses not to define, the transformer will be updated so the cronjob is set to be suspended. This allows the user to manually trigger the job when needed.

Behavior for other scenarios:
- If both are undefined the transformer is not included at all
- If both are defined they will have the cronjob expressions set as the value and .spec.suspend is false.

### Tests
Tested out the new scenario by setting `V4_CFG_VIYA_STOP_SCHEDULE` but not `V4_CFG_VIYA_START_SCHEDULE`. The stop job automatically triggered at the scheduled time and the start did not. I was able to manually trigger the start job.


riragh (approved):
LGTM

dhoucgitter (approved):
lgtm

thpang (approved):
LGTM

PR #298, opened 2022-09-16
feat: (IAC-667) add the ability to install the aws ebs csi driver in DAC

dhoucgitter:
Adding support for DAC to install the aws ebs csi driver.
AWS EKS v1.23 and later requires the installation and configuration of the aws elastic block storage csi driver to enable the dynamic provisioning of physical volumes for applications that are configured to use the default storage class.

### DIFF
--- a/docs/CONFIG-VARS.md
+++ b/docs/CONFIG-VARS.md
@@ -315,7 +315,6 @@ The EBS CSI driver is currently only used for kubernetes v1.23 or later AWS EKS 
 
 | Name | Description | Type | Default | Required | Notes | Tasks |
 | :--- | ---: | ---: | ---: | ---: | ---: | ---: |
-| EBS_CSI_DRIVER_ENABLED | Whether to deploy the aws ebs csi driver into the cluster using helm | bool | true | false | | baseline |
 | EBS_CSI_DRIVER_CHART_URL | aws ebs csi driver helm chart url | string | https://kubernetes-sigs.github.io/aws-ebs-csi-driver | false | | baseline |
 | EBS_CSI_DRIVER_CHART_NAME| aws ebs csi driver helm chart name | string | aws-ebs-csi-driver | false | | baseline |
 | EBS_CSI_DRIVER_CHART_VERSION | aws ebs csi driver helm chart version | string | 2.11.1 | false | | baseline |
--- a/docs/CONFIG-VARS.md
+++ b/docs/CONFIG-VARS.md
@@ -309,6 +309,20 @@ Cluster-autoscaler is currently only used for AWS EKS clusters. GCP GKE and Azur
 | CLUSTER_AUTOSCALER_ACCOUNT | cluster autoscaler aws role arn | string | | false | Required to enable cluster-autoscaler on AWS | baseline |
 | CLUSTER_AUTOSCALER_LOCATION | aws region where kubernetes cluster resides | string | us-east-1 | false | | baseline |
 
+### EBS CSI Driver
+
+The EBS CSI driver is currently only used for kubernetes v1.23 or later AWS EKS clusters. 
+
+| Name | Description | Type | Default | Required | Notes | Tasks |
+| :--- | ---: | ---: | ---: | ---: | ---: | ---: |
+| EBS_CSI_DRIVER_ENABLED | Whether to deploy the aws ebs csi driver into the cluster using helm | bool | true | false | | baseline |
+| EBS_CSI_DRIVER_CHART_URL | aws ebs csi driver helm chart url | string | https://kubernetes-sigs.github.io/aws-ebs-csi-driver | false | | baseline |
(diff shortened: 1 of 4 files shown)


docs/CONFIG-VARS.md
```diff
@@ -309,6 +309,20 @@ Cluster-autoscaler is currently only used for AWS EKS clusters. GCP GKE and Azur
 | CLUSTER_AUTOSCALER_ACCOUNT | cluster autoscaler aws role arn | string | | false | Required to enable cluster-autoscaler on AWS | baseline |
 | CLUSTER_AUTOSCALER_LOCATION | aws region where kubernetes cluster resides | string | us-east-1 | false | | baseline |
 
+### EBS CSI Driver
+
+The EBS CSI driver is currently only used for kubernetes v1.23 or later AWS EKS clusters. 
+
+| Name | Description | Type | Default | Required | Notes | Tasks |
+| :--- | ---: | ---: | ---: | ---: | ---: | ---: |
+| EBS_CSI_DRIVER_ENABLED | Whether to deploy the aws ebs csi driver into the cluster using helm | bool | true | false | | baseline |
```
thpang, line 318: This should be removed.

roles/baseline/defaults/main.yml
```diff
@@ -121,6 +121,23 @@ CLUSTER_AUTOSCALER_CONFIG:
       annotations:
         "eks.amazonaws.com/role-arn": "{{ CLUSTER_AUTOSCALER_ACCOUNT }}"
 
+## EBS CSI Driver
+EBS_CSI_DRIVER_ENABLED: true
```
thpang, line 125: Not needed

roles/common/tasks/main.yaml
```diff
@@ -88,6 +88,24 @@
...
+        - tfstate.ebs_csi_account.value|length > 0
+    - name: tfstate - ebs csi driver location
+      set_fact: 
+        EBS_CSI_DRIVER_LOCATION: "{{ tfstate.location.value }}"
+      when:
+        - tfstate.location is defined
+        - tfstate.location.value|length > 0
+    - name: tfstate - k8s server version
+      set_fact: 
+        K8S_SERVER_VERSION: "{{ tfstate.k8s_server_version.value }}"
```
thpang, line 105: Why `k8s_server_version`? This is the kubernetes version. `k8s_version`

roles/common/tasks/main.yaml
```diff
@@ -88,6 +88,24 @@
...
+      when:
+        - tfstate.k8s_server_version is defined
+        - tfstate.k8s_server_version.value|length > 0
```
thpang, line 108: Change `k8s_server_version` -> `k8s_version` Thx.

roles/baseline/defaults/main.yml
dhoucgitter: Doesn't this option, that will default to true, provide the only easy way for a DAC customer deploying to k8s_version 1.23 or later to prevent ebs csi driver installation if they were to need that for some reason? I don't see how it can be a liability if for only that reason. I added it to provide similar behavior to the autoscaling which is always installed but provides that var as a way to prevent it if needed. Just wanted to bring that up for discussion.

roles/common/tasks/main.yaml
dhoucgitter: sure, I'll update that

roles/common/tasks/main.yaml
dhoucgitter: here too

roles/baseline/defaults/main.yml
thpang: As we talked the EBS driver is ALWAYS needed to provide a working `default` storage class. The objective here it to create a functioning cluster out of the box. If someone sets the flag to false we now have to walk them through setting up the driver, etc. themselves. In the case of autoscaling if that flag is set to `false` the cluster still works. Not the case with the EBS driver. If you want to hide the flag, i.e. remove it from the CONFIG_VARS.md file. I am fine with that. Then it's there not documented and we can do some testing if needed. Again, it's more a distraction at this point. AWS says you need the driver for k8s 1.23 or things won't work, so we provide it.

dhoucgitter: I will remove it from CONFIG_VARS.md and can see the merits of doing that based on those reasons.

docs/CONFIG-VARS.md
dhoucgitter: Removed in latest update

roles/common/tasks/main.yaml
dhoucgitter: updated to k8s_version as per review comment

roles/common/tasks/main.yaml
dhoucgitter: updated here also

jarpat (approved):

riragh (approved):
LGTM

thpang (approved):
LGTM

PR #303, opened 2022-09-29
feat: (IAC-644) Add DaC support for crunchy v5 

dhoucgitter:
# Changes:
- DaC changes related to supporting internal v5 crunchy postgres db.
- Crunchy v5 support for CDS Postgres is still to be added.
- (IAC-724) Update max_connections for Crunchy v5 internal postgres to support Multi-tenancy
- DaC changes related to supporting external crunchy postgres DBs

# Tests:
See internal tickets for details

PR #307, opened 2022-10-10
feat: (IAC-739) Support a CDSPostgreSQL instance for Crunchy v5

dhoucgitter:
### Changes
Viya Retail and Planning application require the installation of a secondary PostgreSQL instance referred to as the Common Data Store Postgres instance.
This updates includes the support to install a secondary internal instance of Postgres named "cds-postgres" and can be triggered by including the following postgres clause in the ansible-vars.yaml file used to deploy Viya.

```yaml
## Postgres
V4_CFG_POSTGRES_SERVERS:
  default:
    internal: true
  cds-postgres:
    internal: true
```
### Tests
Testing included deploying a fast 2020 Inventory Fulfillment Viya order while specifying two Postgres instances as above. The presence of planning application schemas created in the CDS Postgres instance were used to verify the successful deployment and configuration of the CDS Postgres instance.

### DIFF
--- a/roles/vdm/tasks/main.yaml
+++ b/roles/vdm/tasks/main.yaml
@@ -161,8 +161,8 @@
     - uninstall
     - update
 
-- name: Include Planning
-  include_tasks: planning.yaml
+- name: Include CDS
+  include_tasks: cds.yaml
   tags:
     - install
     - uninstall
--- a/roles/vdm/tasks/main.yaml
+++ b/roles/vdm/tasks/main.yaml
@@ -161,6 +161,13 @@
     - uninstall
     - update
 
+- name: Include Planning
+  include_tasks: planning.yaml
+  tags:
+    - install
+    - uninstall
+    - update
+
 - name: Include Storage
   include_tasks: storage.yaml
   tags:
--- a/roles/vdm/tasks/planning.yaml
+++ b/roles/vdm/tasks/planning.yaml
@@ -1,0 +1,24 @@
+---
+- name: planning - folder check
+  stat:
+    path: "{{ DEPLOY_DIR }}/sas-bases/overlays/sas-planning"
+  register: result
+  tags:
+    - install
+    - uninstall
+    - update
+
+- name: planning - add overlays
+  overlay_facts:
+    cadence_name: "{{ V4_CFG_CADENCE_NAME }}"
+    cadence_number: "{{ V4_CFG_CADENCE_VERSION }}"
+    existing: "{{ vdm_overlays }}"
+    add:
+      - { resources: "overlays/sas-planning", min: "2022.10" }
+      - { transformers: "overlays/sas-planning/sas-planning-transformer.yaml", min: "2022.10" }
+  when:
+    - result.stat.exists
+  tags:
+    - install
+    - uninstall
+    - update
--- a/roles/vdm/tasks/postgres/postgres.yaml
+++ b/roles/vdm/tasks/postgres/postgres.yaml
@@ -5,6 +5,29 @@
     - uninstall
     - update
 
+- name: both internal
+  block:
+    - set_fact:
+        default_postgres_internal: "{{ V4_CFG_POSTGRES_SERVERS.default.internal }}"
+        cds_postgres_internal: "{{ V4_CFG_POSTGRES_SERVERS['cds-postgres'].internal }}"
+      when:
+        - V4_CFG_POSTGRES_SERVERS.default is defined
+        - V4_CFG_POSTGRES_SERVERS['cds-postgres'] is defined
+        - V4_CFG_POSTGRES_SERVERS.default.internal is defined
+        - V4_CFG_POSTGRES_SERVERS['cds-postgres'].internal is defined
+    - fail:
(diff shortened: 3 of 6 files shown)


riragh:
@dhoucgitter There is stray TODO in the `postgres-instance.yaml` file, I think this was added in last round to fix with CDS changes could you please take a look. https://github.com/sassoftware/viya4-deployment/blob/feat-iac-739/roles/vdm/tasks/postgres/postgres-instance.yaml#L156

dhoucgitter:
@riragh, it looks like Jay might have added that note with merge in PR #303 to remind us to re-examine the internal crd and external crd sections. It seems that both of those CRD blocks apply to pre 2022.10 cadence orders looking at the when clause content. I suspect we can make those where clauses more similar while leaving the internal/external comparisons in place.

roles/vdm/tasks/postgres/postgres.yaml
```diff
@@ -5,6 +5,29 @@
     - uninstall
     - update
 
+- name: both internal
+  block:
+    - set_fact:
+        default_postgres_internal: "{{ V4_CFG_POSTGRES_SERVERS.default.internal }}"
```
thpang, line 11: The looping here should be over all postgres_servers. If any one is external then we have to determine how we handle the case of the cds internal database server. A customer may have the retail product in their order but may not deploy that product. Does that keep them from creating a cluster and then deploying other products? The customer will wonder why. Should use the `with_items` block for `for` block to cycle of all V4_CFG_POSTGRES_SERVERS.

thpang (changes requested):

roles/vdm/tasks/postgres/postgres.yaml
thpang: This code can be removed. There is already a check for this.

riragh: Ran through couple scenarios with default and CDS postgres, this code block verifies that cds-postgres server is internal if not fails. 
Keeping this block, as it safeguards the cds-postgres internal condition check for both internal and external servers.

thpang: The code here simply needs to check if any of the `V4_CFG_POSTGERS_SERVERS` is external if the cds database is present. Should be much simpler.

thpang (changes requested):

roles/vdm/tasks/main.yaml
```diff
@@ -161,6 +161,13 @@
     - uninstall
     - update
 
+- name: Include Planning
+  include_tasks: planning.yaml
```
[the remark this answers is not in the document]
riragh: Renamed the planning file and tasks to cds.

roles/vdm/tasks/planning.yaml
```diff
@@ -0,0 +1,24 @@
+---
+- name: planning - folder check
```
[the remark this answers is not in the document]
riragh: Renamed the task.

roles/vdm/tasks/postgres/postgres.yaml
riragh: Simplified the check and verified the following scenarios with the new code:
**External postgres servers configured:**

| **default postgres** | **cds-postgres** | **Error message**                                             |
| ---------------- | ------------ | --------------------------------------------------------- |
| False            | True         | Yes. Mix and match error                                  |
| True             | False        | Yes. New error msg                                        |
| True             | True         | No. Successful deployment. Internal databases configured. |
| False            | False        | Yes. New error msg                                        |

**Internal database setup. No external postgres server configured:**

| **default postgres** | **cds-postgres** | **Error message**                                                  |
| ---------------- | ------------ | -------------------------------------------------------------- |
| False            | False        | Yes. New error msg                                             |
| True             | False        | Yes. New error msg                                             |
| False            | True         | Yes. Invalid configuration. Default is expected to be internal |
| True             | True         | No. Successful deployment                                      |

roles/vdm/tasks/postgres/postgres.yaml
```diff
@@ -5,6 +5,19 @@
     - uninstall
     - update
 
+- name: fail - if cds-postgres is external
+  ansible.builtin.fail:
+    msg: "CDS postgres instance must be internal server. To use CDS postgres, please reconfigure your infrastructure to set all your postgres servers internal"
```
thpang, line 10: We might also want to put the link to the SAS documentation stating this limitation : https://go.documentation.sas.com/doc/en/sasadmincdc/v_032/itopssr/p05lfgkwib3zxbn1t6nyihexp12n.htm#n03wzanutmc6gon1val5fykas9aa

thpang (approved):
LGTM. I believe adding the link to the SAS documentation around this limitation would be helpful, but no necessary.

jarpat (approved):

roles/vdm/tasks/postgres/postgres.yaml
riragh: CONFIG-VARS.md has the link mention the details. I added the link you mentioned for additional information on CDS limitation.

PR #310, opened 2022-10-12
fix: (IAC-742) Lookup cluster K8S version information from within DAC

dhoucgitter:
### Changes
Instead of pulling the K8S server version information from an output passed in from tfstate, query the K8s server version from within DAC to remove a dependency on IAC to provide it in the AWS provider, or other providers in the future. This change should allow viya4-iac-aws releases prior to 5.2.0 to work with the latest DAC release just as before.

### Tests
Verification of this update will be run against clusters created by each IAC project.
Internal ticket comments contain dev test results with this update run against clusters created by all four IAC projects.

### DIFF
--- a/roles/baseline/tasks/main.yaml
+++ b/roles/baseline/tasks/main.yaml
@@ -47,7 +47,7 @@
 
 - name: Lookup K8s version info
   block:
-    - name: Retreive K8s cluster information
+    - name: Retrieve K8s cluster information
       community.kubernetes.k8s_cluster_info:
         kubeconfig: "{{ KUBECONFIG }}"
       register: cluster_info
--- a/roles/baseline/tasks/main.yaml
+++ b/roles/baseline/tasks/main.yaml
@@ -53,10 +53,7 @@
       register: cluster_info
     - name: Set the K8s server version
       set_fact:
-        K8S_VERSION: "{{ cluster_info.version.server.kubernetes.major + '.' + cluster_info.version.server.kubernetes.minor }}"
-    - name: Output K8S_VERSION
-      debug:
-        msg: "{{ K8S_VERSION }}"
+        K8S_VERSION: "{{ cluster_info.version.server.kubernetes.major + '.' + cluster_info.version.server.kubernetes.minor | regex_replace('\\+$', '') }}"
   tags:
     - baseline
 
--- a/roles/baseline/tasks/main.yaml
+++ b/roles/baseline/tasks/main.yaml
@@ -45,7 +45,7 @@
   tags:
     - baseline
 
-- name: Query K8s cluster info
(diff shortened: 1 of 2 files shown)


roles/baseline/tasks/main.yaml
```diff
@@ -45,6 +45,21 @@
...
+    - name: Retreive K8s cluster information
+      community.kubernetes.k8s_cluster_info:
+        kubeconfig: "{{ KUBECONFIG }}"
+      register: cluster_info
+    - name: Set the K8s server version
+      set_fact:
+        K8S_VERSION: "{{ cluster_info.version.server.kubernetes.major + '.' + cluster_info.version.server.kubernetes.minor }}"
+    - name: Output K8S_VERSION
+      debug:
+        msg: "{{ K8S_VERSION }}"
```
thpang, line 59: Not really needed here? Also is this generic across all cloud providers?

roles/baseline/tasks/main.yaml
```diff
@@ -45,6 +45,21 @@
...
+  tags:
+    - baseline
```
thpang, line 61: With this line the K8S_VERSION is only used in the `baseline` task during any action `install/uninstall` and not with any other task is that correct?

thpang (changes requested):

roles/baseline/tasks/main.yaml
dhoucgitter: I can remove it if not wanted, I had seen that pattern used for some other set_facts in DAC, just thought it might be useful data in DAC output. This change will be tried with each of the cloud providers prior to asking for final approval based on your earlier suggestion.

roles/baseline/tasks/main.yaml
dhoucgitter: That's the idea yes, right now that var is only referenced to make a decision for including the ebs-csi-driver task during baseline activity.

roles/baseline/tasks/main.yaml
thpang: The fact is only used by ansible so it's not needed for anyone to see. Please remove.

thpang (changes requested):

dhoucgitter:
Debug output removed and a string filter was added to trim the trailing plus sign that showed up when tested against the AWS cluster.

roles/baseline/tasks/main.yaml
dhoucgitter: removed it.

jarpat (approved):

thpang (approved):

riragh (approved):
LGTM

dhoucgitter:
Expect to hold off on merging this until after the 5.6.0 release.

PR #311, opened 2022-10-14
feat: (IAC-754) Allow for extra external DBs & Configurable External DB SSL

jarpat:
### Changes
To allow for parity with Crunchy 4 (<= 2022.09) this change was made to support adding additional PG servers to the deployment in Crunchy 5
Also updates the roles/vdm/templates/transformers/dataserver-transformer.yaml file which is used for the default external PG server so that SSL can correctly be configured.

### Tests
See attached artifacts  and additional details in the internal ticket

Scenarios 1-4 test support for multiple PG servers
Scenarios 5-8 test the update for SSL configurability in the dataserver-transformer.yaml template file

|Scenario|Provider|K8s Version|PG Version|Order|Cadence|Deployment Stabilized|Notes|
| --- | --- | --- | --- | --- | --- | --- | --- |
|1|GCP|v1.23.8-gke.1900|13.7|*|Fast:2020|Yes|Both servers listed under DataServers|
|2|Azure|v1.23.8|13.7|*|Fast:2020|Yes|Both servers listed under DataServers|
|3|OSS|v1.23.8|13.8|*|Fast:2020|Yes|Both servers listed under DataServers|
|4|AWS|v1.23|13.7|*|Fast:2020|Yes|Both servers listed under DataServers|
|5|OSS|v1.23.8|13.8|*|Fast:2020|Yes|server_ssl=off|
|6|GCP|v1.23.8-gke.1900|13.7|*|Fast:2020|Yes|ssl*enforcement*enabled=false|
|7|Azure|v1.23.8|13.7|*|Fast:2020|Yes|2 servers, both ssl*enforcement*enabled=false|
|8|AWS|v1.23|13.7|*|Fast:2020|Yes|ssl*enforcement*enabled=false|


roles/vdm/templates/resources/platform-postgres-dataserver.yaml
```diff
@@ -0,0 +1,35 @@
+{%- set db = db_default_name_map[role] if ('database' not in settings and role in db_default_name_map) else settings.database|default(role, true) -%}
+{%- set server_name = role if role != "default" else "postgres" -%}
```
riragh, line 2: Why this condition based on role? Isn't this file only used in case of `role != "default"`? I am guessing role will never be default, am I missing something?
https://github.com/sassoftware/viya4-deployment/blob/7f7d457459d522f20b5c21fd38ef3e62ec788196/roles/vdm/tasks/postgres/postgres-instance.yaml#L147

roles/vdm/templates/resources/platform-postgres-dataserver.yaml
```diff
@@ -0,0 +1,35 @@
...
+kind: DataServer
+metadata:
+  name: {{ role }}-platform-postgres
+  annotations:
+    sas.com/default-database: "false"
+spec:
+  ssl: {{ (settings.ssl_enforcement_enabled if 'ssl_enforcement_enabled' in settings else True)|bool|lower }}
+  registrations:
+    - serviceName: {{ role }}-sas-postgres
+      dataServiceName: {{ role }}-postgres
```
riragh, line 13: For role = cds-postgres wouldn't this result in `cds-postgres-postgres`, is this expected name?

jarpat: CDS postgres is not supported on external postgres servers. This is adding additional platform-postgres servers.
If you have a deployment directory on hand, it's based off of `/sas-bases/overlays/postgres/platform-postgres/platform-postgres-dataserver.yaml`
This mimics the behavior with Crunchy 4 with this template + multiple external PGs https://github.com/sassoftware/viya4-deployment/blob/main/roles/vdm/templates/resources/postgres-instance.yaml

roles/vdm/templates/resources/platform-postgres-dataserver.yaml
jarpat: I was reusing the working logic from https://github.com/sassoftware/viya4-deployment/blob/main/roles/vdm/templates/resources/postgres-instance.yaml which was from Crunchy 4, I could probably update that to `{%- set server_name = role -%}`

riragh (approved):

roles/vdm/templates/resources/platform-postgres-dataserver.yaml
jarpat: Marking as resolved, this change was put in.

thpang (approved):
LGTM

PR #312, opened 2022-10-16
feat: (IAC-756): Update crunchy v5 tuning transformer

riragh:
# Changes:
- Updated crunchy v5 tuning transformer. The variables and format of the tuning transformer used previously to add max_connections changed, made changes to accommodate these changes.
- Minor change to onboard run command for parity with other commands (this was somehow missed previously).

# Tests:
Verified the new tuning transformers are generated correctly for both platform postgres and cds postgres.


sayeun (approved):
The generated max_connection for different number of tenants looked correct.  👍

PR #313, opened 2022-10-18
fix: (IAC-758) [DAC] error message when uninstalling Viya w/o internal PG

dhoucgitter:
### Changes
Only run the uninstall postgrescluster task when an internal postgres database is present to avoid the error msg generated trying to uninstall a postgres db that's not present.

### Tests
Ran through a Viya uninstall against two deployments, one with an internal postgres db installed and one with an external postgres db installed.

|Scenario |Provider|Outcome|
|--|--|--|
|1 - Internal postgres db|Azure|Uninstall postgresclusters task runs as needed and uninstalls postgres prior to Viya uninstall|
|2- External postgres db|Azure|Uninstall postgresclusters task is skipped|


riragh (approved):

jarpat (approved):
👍

thpang (approved):

PR #316, opened 2022-10-20
(IAC-739) Add auto handling for planning overlay in 2022.10 stable and beyond

dhoucgitter:
With Viya 2022.10 stable releases and later,  DAC has started adding sas-planning overlays automatically if an order has sas-planning.


dhoucgitter:
Abandoning this doc change, PR #318 has removed the need for it.

PR #319, opened 2022-10-20
docs: (IAC-761): Quote the V4_CFG_CADENCE_VERSION value 

riragh:
# Changes:
Without a quoted value for `V4_CFG_CADENCE_VERSION` the YAML parser strips the trailing zeros from the version. As per the [YAML syntax documentation](https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html), If we need to specify a version number, then we will need to quote the value if it looks like a floating-point value.

# Tests:
Verified deployment is successful on passing a quoted value for V4_CFG_CADENCE_VERSION 
`V4_CFG_CADENCE_VERSION: "2022.10"`


thpang (approved):
LGTM

jarpat (approved):

dhoucgitter (approved):
I'd even go so far as to say the the string value Must be quoted so that the version value is interpreted correctly.

docs/CONFIG-VARS.md
```diff
@@ -104,7 +104,7 @@ When setting V4_CFG_MANAGE_STORAGE to true, A new storage classes will be create
 | :--- | ---: | ---: | ---: | ---: | ---: | ---: |
 | V4_CFG_ORDER_NUMBER | SAS order number | string | | true | | viya |
 | V4_CFG_CADENCE_NAME | Cadence name | string | lts | false | [stable,lts] | viya |
-| V4_CFG_CADENCE_VERSION | Cadence version | string | 2020.1 | true | | viya |
+| V4_CFG_CADENCE_VERSION | Cadence version | string | "2020.1" | true | Please quote the string to specify the version correctly | viya |
```
thpang, line 107: For the note field:

```text
This value must be quoted to accommodate the updated SAS Cadence Version format. If the value is not quoted the deployment will fail.
```

riragh: Update the note with provided sample.

dhoucgitter (approved):
lgtm

docs/CONFIG-VARS.md
riragh: Did you actually mean to put a single quote in there, sorry little confused by your suggestion?
> The string must be surrounded by quotation marks`"` to accommodate the updated SAS Cadence Version format.

thpang (approved):

docs/CONFIG-VARS.md
sayeun: Now I look at it, maybe we should say double quotation marks.

riragh: It doesn't matter either is fine, I guess specifying double will just add more questions as why not single.

thpang (approved):

docs/CONFIG-VARS.md
riragh: Final version:
> This value must be surrounded by quotation marks to accommodate the updated SAS Cadence Version format. If the value is not quoted the deployment will fail.

PR #327, opened 2022-11-01
fix: (IAC-742) Lookup cluster K8S version information from within DAC

jarpat:
### Changes
Instead of pulling the K8S server version information from an output passed in from tfstate, query the K8s server version from within DAC to remove a dependency on IAC to provide it in the AWS provider, or other providers in the future. This change should allow viya4-iac-aws releases prior to 5.2.0 to work with the latest DAC release just as before.

### Tests
Verification of this update will be run against clusters created by each IAC project.
Internal ticket comments contain dev test results with this update run against clusters created by all four IAC projects.

Note: merging this back into staging after reverting DO changes.

### DIFF
--- a/roles/baseline/tasks/main.yaml
+++ b/roles/baseline/tasks/main.yaml
@@ -47,7 +47,7 @@
 
 - name: Lookup K8s version info
   block:
-    - name: Retreive K8s cluster information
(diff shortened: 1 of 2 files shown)


riragh (approved):

sayeun (approved):
👍

PR #329, opened 2022-11-01
feat: (IAC-701) Add support for crunchy-storage-transformer and configuration vars, update postgres-storage-transfomer

dhoucgitter:
## Changes
Add support for applying the updated crunchy storage transformer and configuring that transformer via postgres server map variables. Add documentation to CONFIG-VARS for postgres PVC related server map variables.

Add change allowing the storage class specified in the v4 postgres-storage-transformer to take effect
by adding the required storage-type to v4 postgres-storage-transformer

## Tests
See internal ticket for a link to development testing verification information.


docs/CONFIG-VARS.md
```diff
@@ -237,10 +237,26 @@ V4_CFG_POSTGRES_SERVERS:
...
-Example:
+Examples:
 
 ```bash
+# Internal server
+V4_CFG_POSTGRES_SERVERS:
+  default:
+    internal: true
+    postgres_storage_size: 10Gi
+    postgres_access_mode: ReadWriteMany
```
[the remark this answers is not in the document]
dhoucgitter: No, the default in their Readme is `ReadWriteOnce`, I put that mode in the example as it was another possible valid mode that could be specified. Would you recommend that I change it to reflect the default value to avoid confusion?

thpang:
Yeah I shared the information from Christian but am adding it here as well. If we need another PR for that work I am fine with that.

Here is the provided information:

[11:24 AM] Thomas Pangborn
Hi David Houck talked with Christian yesterday and he provided the following information regarding what to look for in the cluster to see if crunchy v4 is there:

--- Christian's comment:

I recommend looking for a k8s deployment named "sas-crunchy-data-postgres-operator". It is specific to v4 and gets pruned automatically (with our kubectl apply -l sas.com/admin=namespace --prune command) if you apply a release with v5

--- My input as to how we leverage existing items in the cluster to remove the need for the flag:

With this information we will not need the new/update flag we can look for that deployment in the cluster and if its there we do not create the storage class but simply use whatever the customer supplies as they had one in play already. If that deployment is not there we create a cluster wide Storage class that ALL Crunchy database transformers will use by default unless the user sets that value.

docs/CONFIG-VARS.md
```diff
@@ -237,10 +237,26 @@ V4_CFG_POSTGRES_SERVERS:
...
+| backrest_storage_size | Size of the internal pgBackrest PVCs | string | 128Gi | false | This value can be changed but not decreased after the initial deployment. Supported for cadence versions 2022.10 and later| viya |
+| postgres_access_mode | Access mode for the PostgreSQL PVCs | string | ReadWriteOnce | false |This value cannot be changed after the initial deployment. Supported for cadence versions 2022.10 and later.| viya |
+| backrest_access_mode | Access mode for the pgBackrest PVCs | string | ReadWriteOnce | false |This value cannot be changed after the initial deployment. Supported for cadence versions 2022.10 and later. | viya |
+| postgres_storage_class | Storage class for the PostgreSQL PVCs | string | default | false |This value cannot be changed after the initial deployment. Supported for cadence versions 2022.10 and later. | viya |
+| backrest_storage_class | Storage class for the pgBackrest PVCs | string | default | false |This value cannot be changed after the initial deployment. Supported for cadence versions 2022.10 and later. | viya |
```
thpang, line 245: All of these items needs a little note that says they are only used for internal databases and not needed for external. Make sure those scenarios are tested.

docs/CONFIG-VARS.md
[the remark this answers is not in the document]
dhoucgitter: Updating the example to reflect the default access mode to avoid suggesting that is preferred.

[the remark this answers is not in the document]
thpang: I don't mind it having the ReadWriteMany in the example to show the differences. You can keep the code or modify.

docs/CONFIG-VARS.md
dhoucgitter: Added "Only for internal databases" note to each.

thpang (approved):
LGTM

docs/CONFIG-VARS.md
```diff
@@ -237,10 +237,26 @@ V4_CFG_POSTGRES_SERVERS:
...
 | connection_name | External postgres database connection name | string | | false | Required for using cloud-sql-proxy on gcp. See [ansible cloud authentication](user/AnsibleCloudAuthentication.md) | viya |
 | service_account | External service account for postgres connectivity | string | | false | Required for using cloud-sql-proxy on gcp. See [ansible cloud authentication](user/AnsibleCloudAuthentication.md) | viya |
+| postgres_storage_size | Size of the internal postgreSQL PVCs | string | 128Gi | false |This value can be changed but not decreased after the initial deployment. Supported for cadence versions 2022.10 and later. Only for internal databases.| viya |
+| backrest_storage_size | Size of the internal pgBackrest PVCs | string | 128Gi | false | This value can be changed but not decreased after the initial deployment. Supported for cadence versions 2022.10 and later. Only for internal databases.| viya |
+| postgres_access_mode | Access mode for the PostgreSQL PVCs | string | ReadWriteOnce | false |This value cannot be changed after the initial deployment. Supported for cadence versions 2022.10 and later. Only for internal databases.| viya |
```
sayeun, line 242: For postgres_access_mode and backrest_access_mode, shouldn’t we show supported values like we did for TLS? Supported values: [ReadWriteOnce,ReadWriteMany].

sayeun: I am going to +2 on this so we can move this forward to testing.  Doc change could happen any time.

sayeun (approved):

PR #332, opened 2022-11-04
(iac-701) Use 'sas' as default storage class name for postgres and pgbackrest PVCs

dhoucgitter:
### Changes
Change the default storage class name to use for postgres and pgbackrest PVCs to 'sas', a storage class  which DAC normally creates for each of the cloud environments.
### Tests
Scenarios run against Azure, AWS and GCP point to the need for this change in AWS and GCP cloud environments.


thpang (approved):
LGTM

jarpat (approved):

riragh (approved):

PR #334, opened 2022-11-09
fix: (iac-788) align V4_CFG_MANAGE_STORAGE default with docs

dhoucgitter:
### Changes
Apply updates/changes requested in the description for IAC-788

Change the default value used for V4_CFG_STORAGE_CLASS to true, that matches the current default expressed in CONFIG-VARS.md.

Update four postgres PVC related configuration variable names per review discussion.

The updates for supported values on lines 242-243 came from review input for IAC-701 that arrived after PR #332 was merged.


docs/CONFIG-VARS.md
```diff
@@ -72,7 +72,7 @@ Tool uses the jump server to interact with rwx filestore, that needs to be pre-m
 
 | Name | Description | Type | Default | Required | Notes | Tasks |
 | :--- | ---: | ---: | ---: | ---: | ---: | ---: |
-| V4_CFG_MANAGE_STORAGE | Should the tool manage the storageclass | bool | true | false | Set to false if you wish to manage the storage class | all |
+| V4_CFG_MANAGE_STORAGE | Should the tool manage the storageclass | bool | false | false | Set to false if you wish to manage the storage class | all |
```
thpang, line 75: this is a pretty big change. Lots of folks depend on this being set to `true`. We'll need a really good reason to break things for folks who have external database and don't care to manager their storage.

dhoucgitter: It's actually not a behavior change but I'd agree with you if it was, it's only a doc update to reflect that the default in our code is really false instead of true. I 
am advocating for only updating the doc to reflect current code behavior and NOT changing the underlying default in the code here ->https://github.com/sassoftware/viya4-deployment/blob/276de04e4e664b59c8ff790a8763998d1e3b2787/roles/vdm/defaults/main.yaml#L28

Our input files in examples/ seem to include "V4_CFG_MANAGE_STORAGE: true" which could result in that variable being set to true often unless they intentionally set it to false.
Check out the comment I made in IAC-788 which argues that updating  the doc is the least disruptive approach of the two so that our doc and code are in sync.

docs/CONFIG-VARS.md
```diff
@@ -239,11 +239,13 @@ V4_CFG_POSTGRES_SERVERS:
...
+| postgres_access_mode | Access mode for the PostgreSQL PVCs | string | ReadWriteOnce | false | Supported values: [`ReadWriteOnce`,`ReadWriteMany`]. This value cannot be changed after the initial deployment. Supported for cadence versions 2022.10 and later. Only for internal databases.| viya |
+| backrest_access_mode | Access mode for the pgBackrest PVCs | string | ReadWriteOnce | false | Supported values: [`ReadWriteOnce`,`ReadWriteMany`]. This value cannot be changed after the initial deployment. Supported for cadence versions 2022.10 and later. Only for internal databases.| viya |
 | postgres_storage_class | Storage class for the PostgreSQL PVCs | string | sas | false |This value cannot be changed after the initial deployment. Supported for cadence versions 2022.10 and later. Only for internal databases.| viya |
 | backrest_storage_class | Storage class for the pgBackrest PVCs | string | sas | false |This value cannot be changed after the initial deployment. Supported for cadence versions 2022.10 and later. Only for internal databases.| viya |
```
thpang, line 245: Wondering if variables need more descriptive names. For example:

```code
postgres_storage_size -> postgres_pv_storage_size
backrest_storage_size -> backreset_pv_storage_size

etc.
```

This then lets the user know this is an item in the kubernetes cluster that they are setting. I know the description explains this but if I am in the code and see `postgres_access_mode` I am guessing that's something to do with the actual database server and not its storage in the cluster. Where `postgres_pv_access_mode` is a bit clearer.

thpang (changes requested):
If the `_pv_` items needs to be `_pvc_` or something a bit different I am fine with that as well.

thpang:
Having a really hard time with this one. It seems we're now forcing customers to understand their storage class type when creating their postgres servers and these seem like two different questions. @dhoucgitter setup a meeting to discuss. Thx

thpang (changes requested):

docs/CONFIG-VARS.md
dhoucgitter: > Having a really hard time with this one. It seems we're now forcing customers to understand their storage class type when creating their postgres servers and these seem like two different questions. @dhoucgitter setup a meeting to discuss. Thx

I'll make an update to the variable name as requested to emphasize the storage aspect and also setting up a discussion with you.

dhoucgitter: @thpang would these name changes for the new postgres vars help emphasize they are PVC and not actual database server items?
I would also make the same name changes for the three vars prefixed with backrest_
```
postgres_storage_size -> postgres_pvc_storage_size
postgres_access_mode -> postgres_pvc_access_mode
postgres_storage_class -> postgres_pvc_storage_class
```

thpang: yes that's the plan. The information in that block is for the database server. The PVC is only for the internal crunchy database. So any name changes to identify why these items are there would be most helpful.

thpang (approved):
LGTM

docs/CONFIG-VARS.md
dhoucgitter: Per team discussion this morning I've restored CONFIG-VARS.md so that the default indicated in the table for V4_CFG_MANAGE_STORAGE is true and I've updated the vdm defaults main.yaml file, line 28 so that the default value for V4_CFG_MANAGE_STORAGE is now true.

riragh (approved):

PR #338, opened 2022-11-16
feat: (iac-770) Create new storage class for crunchy v5 PVCs

dhoucgitter:
### Changes

- Create the `pg-storage` storage class with `reclaimPolicy: retain` during the baseline install task
- By default, use the `pg-storage` storage class for internal postgres database server instances.
- Once a set of postgres_servers variable map storage variables have been applied to a cluster during a deployment, the storage class and access modes for those servers can not be updated with a DAC deployment to the same cluster without uninstalling Viya.
- Switch from using the `community.kubernetes` ansible collection modules to the `kubernetes.core` ansible collection modules, update Dependencies.md and requirements.yaml

### Tests
Development testing to include OOTB 2022.10 executions of DAC to generate site-config/vdm content using `DEPLOY:false`. Once `DEPLOY:true` has deployed Viya to the cluster, verify that no updates to v5 crunchy storage class or access mode for either default or cds-postgres instances is allowed along with the output of an ansible message providing that notice to the DAC user.
See internal ticket for additional test details including external postgres db, 2022.09 stable order and that the new storage class works properly when "ha" storage_type is configured for each cloud environment.


roles/baseline/defaults/main.yml
```diff
@@ -90,6 +90,26 @@ NFS_CLIENT_CONFIG:
     archiveOnDelete: "false"
     name: sas
 
+## pg-storage storage class config
+PG_NFS_CLIENT_NAME: nfs-subdir-external-provisioner-pg-storage
```
jarpat, line 94: Need to update the names in `CONFIG-VARS.md` as well

jarpat: Nevermind didn't realize this were new variables, and the old ones still exist. Do we need to doc these?

dhoucgitter: I will add doc to include the new PG_ set of variables in `CONFIG-VARS.md` and update the default for NFS_CLIENT_NAME as well.

jarpat (approved):

riragh (approved):
LGTM

roles/vdm/tasks/postgres/pg_config.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - name: set fact that ConfigMap was found
+    set_fact:
+      dac_crunchy_storage_cm_found: true
+    ignore_errors: true
+    when: "cm_info.resources"
+  - name: Use previously applied postgres storage map values
+    debug: 
+      msg:
+      - "The previous internal postgres_servers variable map values for storage class and access mode will be used."
+      - "The current internal postgres_servers variable map for storage class and access mode values will be ignored!"
```
thpang, line 23: I know what you're trying to say here, but it's a bit muddled. You might just want to say the previously found value for the postgres server storage will be used and that the values you have in your ansible vars will be ignored because the storage class is already present in your cluster.

thpang (changes requested):
Items to review and comments

roles/vdm/tasks/postgres/pg_config.yaml
```diff
@@ -0,0 +1,50 @@
...
+    register: find_op
+    ignore_errors: true
+  - name: v4 crunchy operator found
+    set_fact:
+      v4_crunchy_present: true
+    when: "find_op.rc == 0"
+  - name: v4 crunchy operator not found
+    set_fact:
+      v4_crunchy_present: false
+    when: "find_op.rc != 0"
```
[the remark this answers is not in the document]
dhoucgitter: Looked into that further and it appears the ansible arg|bool filter parses it's argument as a string looking for 'yes, on, true, false, False, Off, NO' and additional words. It does not appear that the ansible bool filter will allow for strict test of rc == 0 or rc != 0 like using when with the equality test does. In other words, it is not a Pythonic test for truthiness as one might assume.

roles/vdm/tasks/postgres/pg_config.yaml
dhoucgitter: Updated to say:
- "The previously found values for the postgres server storage class and access mode will be used."
- "The values you have in your ansible vars will be ignored because the storage class is already present in your cluster."

roles/vdm/tasks/postgres/postgres-instance.yaml
```diff
@@ -134,6 +170,69 @@
...
+              namespace: "{{ NAMESPACE }}"
+            data:
+              cds-postgres-storage-class: "{{ cds_postgres_storage_class | string }}"
+              cds-postgres-access-mode: "{{ cds_postgres_pvc_access_mode | string }}"
+              cds-backrest-storage-class: "{{ cds_backrest_storage_class | string }}"
+              cds-backrest-access-mode: "{{ cds_backrest_pvc_access_mode | string }}"
+      when: 
+      - not dac_crunchy_storage_cm_found
+      - DEPLOY
+      - role == "cds-postgres"
```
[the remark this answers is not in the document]
dhoucgitter: yep, updated this and 2 other locations to be consistent, found out that yaml lists aren't req'd to be indented, just all at the same level.

roles/vdm/tasks/postgres/postgres-instance.yaml
```diff
@@ -134,6 +163,69 @@
...
+              name: dac-crunchy-storage-info
+              namespace: "{{ NAMESPACE }}"
+            data:
+              default-postgres-storage-class: "{{ default_postgres_storage_class | string }}"
+              default-postgres-access-mode: "{{ default_postgres_pvc_access_mode | string }}"
+              default-backrest-storage-class: "{{ default_backrest_storage_class | string }}"
+              default-backrest-access-mode: "{{ default_backrest_pvc_access_mode | string }}"
+      when: 
+      - not dac_crunchy_storage_cm_found
+      - DEPLOY
```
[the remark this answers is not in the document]
AWSmith0216: I haven't reviewed the entire intent of this merge, but if you're creating a ConfigMap, would such a config map be pruned out during subsequent deployments? The 'kubectl apply' (and deployment operator) used during deployment includes prune options that may unexpectedly remove what you create.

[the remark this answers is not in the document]
dhoucgitter: @AWSmith0216, I have not seen that yet in the tests run so far with orchestration tooling, the ConfigMap has not been pruned out with subsequent deployments. I'll verify that the same thing is true with subsequent deployments using the deployment operator.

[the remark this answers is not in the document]
AWSmith0216: The orchestration tooling should behave the same as the operator I believe, so if you've confirmed that then you're probably good. Just wanted to mention it to make sure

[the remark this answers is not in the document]
dhoucgitter: I ran through a development test running multiple deployments with the deployment operator and can confirm that the ConfigMap DAC creates is not pruned during a second subsequent deployment.

roles/vdm/tasks/postgres/pg_config.yaml
```diff
@@ -0,0 +1,63 @@
...
+      name: dac-crunchy-storage-info
+      namespace: "{{ NAMESPACE }}"
+    register: cm_info
+  - name: print postgres_servers ConfigMap values
+    debug:
+      msg: "{{ cm_info.resources[0].data }}"
+    when: "cm_info.resources"
+  - name: print one ConfigMap data value
+    debug:
+      msg: "{{ cm_info.resources[0].data['v4-crunchy-operator-found'] }}"
```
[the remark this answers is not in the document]
thpang: I don't see these comments in the updated code or a comment from @dhoucgitter on this to know if its been addressed.

roles/vdm/tasks/postgres/pg_config.yaml
```diff
@@ -0,0 +1,63 @@
...
+    when: "cm_info.resources"
+  - name: set facts from configMap keys
+    set_fact:
+      dac_crunchy_storage_cm_found: true
+      v4_crunchy_present: "{{ cm_info.resources[0].data['v4-crunchy-operator-found'] }}"
```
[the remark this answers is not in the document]
thpang: I don't see these comments in the updated code or a comment from @dhoucgitter on this to know if its been addressed.

thpang (approved):
LGTM

roles/vdm/tasks/postgres/postgres-instance.yaml
```diff
@@ -122,6 +122,45 @@
 
 - name: postgres instance - internal post 2022.10
   block:
+  - name: postgres instance - crunchy v4 to crunchy v5 update is required
+    fail:
+      msg: "A manual update from crunchy v4 to v5 is required, follow these steps https://pubshelpcenter.unx.sas.com/test/doc/en/itopscdc/v_033/dplynotes/p1bpcvd3sr8au8n1w9ypcvu31taj.htm"
```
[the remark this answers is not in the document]
dhoucgitter: @riragh, good catch. pushed update to use the external link.

roles/vdm/tasks/postgres/pg_config.yaml
```diff
@@ -0,0 +1,57 @@
+- name: pg_config - Check for storage info ConfigMap
+  block:
+  - name: set facts 
+    set_fact:
+      dac_crunchy_storage_cm_found: false
+      v4_crunchy_present: false
```
thpang, line 6: We have both `found` and `present` in 2 facts. Lets just use `found` as the flag item. Also wondering why the `v4_crunchy_present` value is here if the logic is looking for the config map about storage.

roles/vdm/tasks/postgres/postgres-instance.yaml
```diff
@@ -122,6 +122,45 @@
 
 - name: postgres instance - internal post 2022.10
   block:
+  - name: postgres instance - crunchy v4 to crunchy v5 update is required
+    fail:
+      msg: "A manual update from crunchy v4 to v5 is required, follow these steps https://go.documentation.sas.com/doc/en/itopscdc/default/dplynotes/p1bpcvd3sr8au8n1w9ypcvu31taj.htm"
+    when:
+      - v4_crunchy_present
```
thpang, line 129: This seems like the more logical place for the `v4_crunchy_present` flag to be defined though the new name would be `v4_crunchy_found`

roles/vdm/tasks/postgres/postgres-instance.yaml
```diff
@@ -134,6 +173,69 @@
...
+            apiVersion: v1
+            kind: ConfigMap
+            metadata:
+              name: dac-crunchy-storage-info
+              namespace: "{{ NAMESPACE }}"
+            data:
+              default-postgres-storage-class: "{{ default_postgres_storage_class | string }}"
+              default-postgres-access-mode: "{{ default_postgres_pvc_access_mode | string }}"
+              default-backrest-storage-class: "{{ default_backrest_storage_class | string }}"
+              default-backrest-access-mode: "{{ default_backrest_pvc_access_mode | string }}"
```
thpang, line 207: This code seems to always update the config map, is that the desired effect here? Thinking we set the values once and then don't set them again ever if we find them.

roles/vdm/tasks/postgres/pg_config.yaml
dhoucgitter: I was setting v4_crunchy_present to false here to define an acceptable default for the fact. I can remove it here and add a when is defined prior to testing for true or false in postgres-instance.

roles/vdm/tasks/postgres/postgres-instance.yaml
dhoucgitter: adding is defined test prior to truthiness test in the when clause, will no longer need the default value set in line 6 of pg_config.yaml above, use v4_crunchy_found for flag name

roles/vdm/tasks/postgres/postgres-instance.yaml
dhoucgitter: It should not update the config map if the config map already exists, the when clause will prevent the config map from being updated when the dac_crunchy_storage_cm_found flag is set to true.

dhoucgitter:
@thpang I've posted updates in response to your review comments, ready for re-review

PR #345, opened 2022-11-21
feat: (IAC-664) Update Docker Run Calls and Create License Folder

jarpat:
## Changes
The addition of "license" folder, this will live underneath the DEPLOY_DIR and it will contain the certs and license.jwt file. The license.jwt currently exists under site-config. This is so the structure is more in line in the SAS Documentation.

Swapped out the usage of `community.docker.docker_container` to run the orchestration docker container to instead use `ansible.builtin.shell`. The older module did not have a way to split stdout from stderr and included lines we did not need in the generated CRD .

## Tests
See internal ticket.
| Scenario | Provider | K8s Version | Order | Docker | Cadence | DO Enabled | Deployment Stabilized  |  
|---|---|---|---|---|---|---|---|
| 1 | Azure | v1.23.8 | * | With Docker | stable:2022.11 | yes | yes  |  
| 2 | GCP | v1.24.4-gke.800 | *| With Docker | stable:2022.11 | no | yes  |  
| 3 | GCP | v1.24.4-gke.800 | *| Without Docker | stable:2022.11 | yes | yes  |  
| 4 | Azure | v1.23.8 | *| Without Docker | stable:2022.11 | no | yes  |

### DIFF
--- a/roles/vdm/tasks/deploy.yaml
+++ b/roles/vdm/tasks/deploy.yaml
@@ -55,22 +55,18 @@
       when:
         - deployment_tooling == "docker"
     - name: deploy - Deploy SAS Viya
-      community.docker.docker_container:
-        name: "orchestration_{{ lookup('password', '/dev/null chars=ascii_lowercase length=8') }}"
-        image: "{{ V4_CFG_CR_HOST }}/{{ ORCHESTRATION_IMAGE }}"
-        volumes:
-          - "{{ ORCHESTRATION_TOOLING_INSTALL_MANIFESTS_DIRECTORY }}:/manifests"
-          - "{{ KUBECONFIG }}:/config/kubeconfig"
-        detach: false
-        cleanup: yes
-        user: "{{ UID_GID }}"
-        env:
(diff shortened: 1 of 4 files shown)


riragh (approved):

roles/vdm/tasks/deploy.yaml
```diff
@@ -55,22 +55,18 @@
       when:
         - deployment_tooling == "docker"
     - name: deploy - Deploy SAS Viya
-      community.docker.docker_container:
```
dhoucgitter, line 58: Are there still uses of community.docker with the switch to builtin.shell? Wondering about removing the references from Dependencies.md and requirements.yaml if not.

jarpat: Yes, there are 2 instances of `community.docker` in `viya4-deployment/public/roles/vdm/tasks/orchestration.yaml`. It's still being used to manage the docker login process as well as initially pulling the orchestration image.

thpang (approved):
LGTM

PR #350, opened 2022-12-02
fix: (IAC-664) Move License Folder & Update Docker Run Call

jarpat:
### Change

Updates after review from testing:
Moved the license folder so it's more in line with documentation. It should actually be "parallel" with the "$deploy" directory.
Because it was moved up a level, the sas-license.yaml secret generator template ran into issues, since the file was no longer contained with the kustomization root, I updated the template to instead take the literal license value rather than the path to avoid this issue.

### Tests

See internal ticket for more details, verified the new license folder location and that deployments would still work.
|Scenario|Provider|K8s Version|Cadence|DO/Deploy|Docker/Ansible|
|:----|:----|:----|:----|:----|:----|
|1|GCP|v1.23.8-gke.1900|fast:2020|DO|Ansible|
|2|Azure|v1.23.8-gke.1900|stable:2022.11|DO|Docker|
|3|GCP|v1.23.8-gke.1900|fast:2020|Deploy|Ansible|
|4|Azure|v1.23.8-gke.1900|stable:2022.11|Deploy|Docker|


dhoucgitter (approved):
LGTM

sayeun (approved):
👍

PR #351, opened 2022-12-06
feat: (IAC-770) use k8s.core.k8s_info module not kubectl to avoid ansible error output

dhoucgitter:
### Changes
Substitute using an ansible kubernetes.core.k8s_info module query as opposed to the builtin.shell "kubectl | grep" command to avoid generating "red" Ansible error output that requires ignoring and can cause questions or concerns for end users.
### Testing
Ran through a deployment with Viya uninstalled and two separate redeployments against clusters with v4 crunchy installed and a redeployment on a cluster with v5 crunchy installed to verify that the v4_crunchy_present fact was being set properly to true or false for the version of crunchy installed.


thpang (approved):
LGTM

riragh (approved):

PR #395, opened 2023-02-20
feat: (IAC-911) Utilize kustomize from within the sas-orchestration image for all tasks

jarpat:
### Changes

To remove the overhead of managing the kustomize version we want to use the kustomize binary that is already included within the sas-orchestration image.

The image that is pulled down (sourced from sas-bases/.orchestration/images.yaml) should always have a version of kustomize packaged with it that is compatible with the cadence that is going to be deployed.

We were already using the image previously for building the CRD, this change moves the image pull a little earlier so we can utilize it to generate the manifest for the DO and use it to generate the MT onboard/offboard manifests.

### Tests

Ran through the following scenarios, see internal ticket for details

|Scenario|Task|Provider|Order|Cadence|K8s Version|Execution Method|Orchestration|
|:----|:----|:----|:----|:----|:----|:----|:----|
|1|OOTB + Multi-Tenant|Azure|*|stable:2023.01|v1.23.12|Docker|Deployment Operator|
|2|OOTB + Multi-Tenant|GCP|*|stable:2023.02|v1.24.9-gke.2000|Ansible|Deploy Command|
|3|OOTB|Azure|*|stable:2023.01|v1.23.12|Docker|Deploy Command|
|4|OOTB|GCP|*|fast:2020|v1.24.9-gke.2000|Ansible|Deployment Operator|


riragh (approved):

anthoday (approved):

roles/multi-tenancy/tasks/multi-tenant-onboard-offboard.yaml
```diff
@@ -52,34 +52,62 @@
...
   block:
-    - name: kustomize - generate onboard manifest
-      ansible.builtin.shell: |
-        kustomize build {{ DEPLOY_DIR }}/site-config/sas-tenant-job/ --load-restrictor LoadRestrictionsNone -o {{ DEPLOY_DIR }}/site-config/sas-tenant-job/site-onboard.yaml
-      when: V4_CFG_CADENCE_VERSION is version('2023.02', ">=") or V4_CFG_CADENCE_NAME|lower == "fast"
-    - name: kustomize - generate onboard manifest - legacy flags
-      ansible.builtin.shell: |
-        kustomize build {{ DEPLOY_DIR }}/site-config/sas-tenant-job/ --load_restrictor='none' -o {{ DEPLOY_DIR }}/site-config/sas-tenant-job/site-onboard.yaml
+    - name: kustomize - generate onboard manifest - Ansible
+      ansible.builtin.shell: >
```
thpang, line 65: What's the purpose of the `>` directive vs the pipe `|` directive?

roles/multi-tenancy/tasks/multi-tenant-onboard-offboard.yaml
```diff
@@ -52,34 +52,62 @@
...
   block:
-    - name: kustomize - generate offboard manifest
-      ansible.builtin.shell: |
-        kustomize build {{ DEPLOY_DIR }}/site-config/sas-tenant-job/ --load-restrictor LoadRestrictionsNone -o {{ DEPLOY_DIR }}/site-config/sas-tenant-job/site-offboard.yaml
-      when: V4_CFG_CADENCE_VERSION is version('2023.02', ">=") or V4_CFG_CADENCE_NAME|lower == "fast"
-    - name: kustomize - generate offboard manifest - legacy flags
-      ansible.builtin.shell: |
-        kustomize build {{ DEPLOY_DIR }}/site-config/sas-tenant-job/ --load_restrictor='none' -o {{ DEPLOY_DIR }}/site-config/sas-tenant-job/site-offboard.yaml
+    - name: kustomize - generate offboard manifest - Ansible
+      ansible.builtin.shell: >
```
thpang, line 91: Same question from above.

roles/vdm/tasks/sas_deployment_cr.yaml
```diff
@@ -1,161 +1,27 @@
+# These tasks are responsible for generating the SASDeployment Custom Resource
+
 ---
-- name: orchestration - Set orchestration tooling facts
+- name: sas deployment cr - Set orchestration tooling facts
```
thpang, line 4: There are several references in the task names to `sas deployment cr` is the the SAS Deployment Container Registry? I know it's not, but the `cr` abbreviation is currently used to stand for just that. If this has another meaning let's find a better acronym. Too many below with the same issue. I'll just leave this one comment.

thpang (changes requested):

roles/multi-tenancy/tasks/multi-tenant-onboard-offboard.yaml
jarpat: I'm using the ">" directive (Folded Block Scalar) so that new lines in the middle of my multiline block are interpreted as spaces and not newlines. If the pipe was used every line where I split up my multiline block would be interpreted as a newlines.

A quick example
```
- name: my task
  ansible.builtin.shell: >
	kubectl get
	pods -n mynamespace
```
would be run as `kubectl get pods -n mynamespace\n`

```
- name: my task
  ansible.builtin.shell: |
	kubectl get
	pods -n mynamespace
```

would be run as `kubectl get\npods -n mynamespace\n`. The newline that is retained in the middle of the command would be problematic when using the pipe.

No direct link to it in the documentation, but search for the string "Folded Block Scalar" on the Ansible doc page. Which has a good explanation 
https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html#yaml-basics

roles/vdm/tasks/sas_deployment_cr.yaml
jarpat: I can see that task name being confusing and it does conflict with the  word "container registry". I think I'll forgo using an abbreviation for the task names and file name and change it to the whole name.

new file name: `roles/vdm/tasks/sasdeployment_custom_resource.yaml`
example task name update: `- name: sasdeployment custom resource- Set orchestration tooling facts`

Lines up better with the SAS documentation here: https://go.documentation.sas.com/doc/en/itopscdc/v_037/dplyml0phy0dkr/p0nid9gu3x2cvln1pzpcxa68tpom.htm

jarpat: Updates made, ready for re-review.

dhoucgitter (approved):

PR #401, opened 2023-03-01
feat: (IAC-892): set default kubectl to 1.24

riragh:
# Changes:
With SAS Viya Platform supporting K8s 1.23, 1.24, and 1.25 in 2023.03 March release cadence, updating our supported kubectl version to 1.24.10 for it to be in the +/- 1 range of the supported versions.

# Tests: WIP

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN apt update && apt upgrade -y \
   && update-alternatives --install /usr/bin/pip pip /usr/bin/pip3 1
 
 FROM baseline as tool_builder
-ARG kubectl_version=1.23.8
+ARG kubectl_version=1.24.10
 
 WORKDIR /build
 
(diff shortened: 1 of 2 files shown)


sayeun:
Don't we need to update https://github.com/sassoftware/viya4-deployment/blob/iac-892/roles/baseline/tasks/main.yaml#L65?

riragh:
> Don't we need to update https://github.com/sassoftware/viya4-deployment/blob/iac-892/roles/baseline/tasks/main.yaml#L65?

No, the task `Include ebs-csi-driver` is still valid for 1.23 and above.

jarpat (approved):

thpang (approved):

sayeun (approved):
Thanks Ritika.  I am good with this.

PR #428, opened 2023-04-12
feat: (IAC-298) Security update - `use-forwarded-headers` set to false for ingress-ngnix

riragh:
# Changes:
Previously the default value for `use-forwarded-headers` was set to true. This has raised a security concern as it allows spoofing source IP via X-Forwarded-For header.

[NGINX Ingress Controller](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#use-forwarded-headers) states the use of `use-forwarded-headers` as follows:

>If true, NGINX passes the incoming X-Forwarded-* headers to upstream. Use this option when NGINX is behind another L7 proxy / load balancer that is setting these headers.
>
>If false, NGINX ignores incoming X-Forwarded-* headers, filling them with the request information it sees. Use this option if NGINX is exposed directly to the internet, or it's behind a L3/packet-based load balancer that doesn't alter the source IP in the packets.

As NGINX is not behind another L7 proxy / load balancer this PR will set the `use-forwarded-headers` to false by default. Instructions are added in troubleshooting guide if user wishes to enable this setting.

# Tests: 
Verified on following scenarios, the deployment stabilized and applications were accessible using the change `use-forwarded-headers = false`. See additional details and tests in internal ticket:
|Scenario|Task|Provider|Cadence|
|:----|:----|:----|:----|
|1|OOTB|Azure|fast:2020|
|2|OOTB|AWS|fast:2020|
|3|OOTB|GCP|fast:2020|

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -8,7 +8,7 @@
   - [kustomize - Generate deployment manifest](#kustomize---generate-deployment-manifest)
   - [Ingress-Nginx issue - Unable to access SAS Viya Platform web apps](#ingress-nginx-issue---unable-to-access-sas-viya-platform-web-apps)
   - [Ansible Variables with Special Jinja2 Characters](#ansible-variables-with-special-jinja2-characters)
-
+  - [Ingress-Nginx - use-forwarded-headers disabled](#ingress-nginx---use-forwarded-headers-disabled)
 
 ## Debug Mode
 Debug mode can be enabled by adding "-vvv" to the end of the docker or ansible commands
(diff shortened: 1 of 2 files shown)


jarpat (approved):

thpang (approved):
I am ok with this as long as we tested the SAS coded and verified that the systems still work without this item.

riragh:
> I am ok with this as long as we tested the SAS coded and verified that the systems still work without this item.

Yes this change was verified on all the cloud providers, SAS Viya Platform deployments were successful and applications were accessible.

dhoucgitter (approved):

PR #433, opened 2023-04-24
feat: (IAC-1002) Upgrade the default ingress-nginx and kubectl versions to work with K8s 1.26

riragh:
# Changes:
SAS Viya Platform is going to support a range of K8s `1.24-1.26` in cadence `v2023.05`. This PR contains following updates:
- The default kubectl version updated to `1.25.8`
- The default ingress-nginx version updated:
  - For K8s version `>= 1.24.X` the default ingress-nginx version is updated to the latest app version: `1.7.0` or chart version: `4.6.0`.
  - For K8s version `<= 1.23.X` the default ingress-nginx version is updated to the latest app version: `1.4.0` or chart version: `4.3.0`. **Note**: This is same version from previous release.

# Tests:
Verified SAS Viya Platform deployment is successful on the following scenarios, see details in internal ticket:

|Scenario|Task|Provider|Cadence|kubernetes_version|Deploy method|
|:----|:----|:----|:----|:----|:----|
|1|OOTB|Azure|fast:2020|1.25|ansible, DO: true|
|2|OOTB|Azure|fast:2020|1.26|ansible, DO: true|
|3|OOTB|Azure|Stable 2023.04|1.24|docker, DO: false|
|4|OOTB|AWS|Stable 2023.04|1.23|docker, DO: true|

### DIFF
--- a/roles/baseline/defaults/main.yml
+++ b/roles/baseline/defaults/main.yml
@@ -31,15 +31,15 @@ METRICS_SERVER_CONFIG:
 ## Ingress-nginx - Defaults
 ingressVersions:
   k8sMinorVersionCeiling:
-    value: 21
-    api:
-      chartVersion: 3.40.0
-      appVersion: 0.50.0
-  k8sMinorVersionFloor:
-    value: 22
+    value: 23
     api:
       chartVersion: 4.3.0
       appVersion: 1.4.0
+  k8sMinorVersionFloor:
+    value: 24
+    api:
+      chartVersion: 4.6.0
+      appVersion: 1.7.0
 
 ## Ingress-nginx - Ingress
 INGRESS_NGINX_NAME: ingress-nginx
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN apt update && apt upgrade -y \
   && update-alternatives --install /usr/bin/pip pip /usr/bin/pip3 1
 
 FROM baseline as tool_builder
-ARG kubectl_version=1.24.10
+ARG kubectl_version=1.25.8
 
 WORKDIR /build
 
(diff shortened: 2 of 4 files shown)


roles/baseline/defaults/main.yml
```diff
@@ -31,15 +31,15 @@ METRICS_SERVER_CONFIG:
...
   k8sMinorVersionCeiling:
-    value: 21
-    api:
-      chartVersion: 3.40.0
-      appVersion: 0.50.0
-  k8sMinorVersionFloor:
-    value: 22
+    value: 23
     api:
       chartVersion: 4.3.0
```
thpang, line 36: I see this is a standard but why do we have the `appVersion` along with the `chartVersion`? The `chartVersion` will control the application version being installed. Seems like extra work.

riragh: Good catch, I agree that `chartVersion` should take care of what application version is being installed, but it's been present in code for a while so I'll check if the `appVersion` is used anywhere else in the code?

jarpat: @thpang the appVersion is not necessarily required and could be removed without consequence.

When this block was initially implemented I added it as documentation so a user/dev could easily lookup the version in the [ingress-kubernetes compatibility table](https://github.com/kubernetes/ingress-nginx#supported-versions-table) without having to jump into the helm index yaml to determine what the app version is.
See this comment: https://github.com/sassoftware/viya4-deployment/pull/181#discussion_r791961151

thpang: I would suggest from a support perspective we remove this in the future.

thpang (approved):

jarpat (approved):
I'm ok with removing it as part of the next ingress version bump

PR #435, opened 2023-04-26
feat: (IAC-696) DAC - Install gke-gcloud-auth-plugin with viya4-deployment for K8s v1.26

dhoucgitter:
# Changes
Existing versions of kubectl and custom Kubernetes clients contain provider-specific code to manage authentication between the client and Google Kubernetes Engine. Starting with v1.26, that authentication code will no longer be included as part of the OSS kubectl. The `gke-gcloud-auth-plugin` will fill the gap for the authentication code that is being removed from kubectl.

viya4-deployment supports using two types of GKE kubernetes configuration files, also known as kubeconfig files. The `gke-gcloud-auth-plugin` binary is required to access a GKE clusters when using kubectl 1.26+ with a "provider based" kubernetes configuration file. The "service account and cluster role binding" kubernetes configuration file variant remains the same and still does not require either `gcloud` or the `gke-gcloud-auth-plugin` binary to communicate with the cluster.
The `gke-gcloud-auth-plugin` binary will now be installed as part of the viya4-deployment docker container and is available to the `gcloud` and kubectl binaries within the container.
If a viya4-deployment user has chosen to use a provider based kubernetes configuration file and runs viya4-deployment via Ansible, they will need to install the `gcloud` binary and `gke-gcloud-auth-plugin` using the documented instructions.
See doc here: https://github.com/sassoftware/viya4-deployment/pull/435/files#diff-8a22ddc8bcdf8b5b392d4e76e637e4e6fd79cba9ebf343dddef54713b947a293R20

### Related viya4-iac-gcp changes
Updates in the [PR](https://github.com/sassoftware/viya4-iac-gcp/pull/169) for viya4-iac-gcp have been made to how the Provider based kubernetes configuration file is generated. The format has been updated to support the use of the `gke-gcloud-auth-plugin`. The `gke-gcloud-auth-plugin` allows client authentication to 1.26 GKE clusters with provider based kubernetes configuration files since provider specific authentication code will not be included with OSS kubectl any longer.

# Tests
|Scenario | Task | Deployment method | Task tags | Security | KUBECONFIG |V4_CFG_CLOUD_SERVICE_ACCOUNT_NAME | gcloud version | gke-gcloud-auth-plugin | kubectl version | k8s version | Orchestration | Provider | Cadence | INGRESS_NGINX_CHART_VERSION |Notes
|:--|:--|:--|:--|:--|:--|:--|:--|:--|:--|:--|:--|:--|:--|:--|:--|
1 | OOTB | docker | baseline,viya,install | front door TLS | non-static kubeconfig (auth-plugin) |sa_name@project_name.iam.gserviceaccount.com | 427.0.0 | 0.5.2 (reported from gcloud version ) | 1.25.8 | 1.26.3 | deploy | GCP | Fast | 4.5.2
2 |  OOTB | docker | baseline,viya,install | same as above | static kubeconfig | n/a | 428.0.0 | 0.5.2 | 1.25.8 | 1.26.3 | deploy | GCP | Fast | 4.5.2
3 |  OOTB | ansible | baseline,viya,install | front door TLS | non-static kubeconfig (auth-plugin) | sa_name@project_name.iam.gserviceaccount.com | 427.0.0 | 0.5.2 | 1.25.5 | 1.26.3 | deploy | GCP | Fast | 4.5.2 | Ansible deployment method with non-static kubeconfig is not supported 
4 | OOTB | ansible | baseline,viya,install | same as above | static kubeconfig | n/a | 428.0.0 | 0.5.2 | 1.25.8 | 1.26.3 | deploy | GCP | Fast | 4.5.2
5 | OOTB | docker | viya, install | front door TLS | non-static kubeconfig (auth-plugin) | sa_name@project_name.iam.gserviceaccount.com | 428.0.0 | 0.5.2 | 1.25.8 | 1.26.3 | DO | GCP | Fast | 4.5.2
6 | logging/ monitoring | docker | baseline,cluster-logging,cluster-monitoring,viya,install | front door TLS | non-static kubeconfig (auth-plugin) | sa_name@project_name.iam.gserviceaccount.com | 428.0.0 | 0.5.2 | 1.25.8 | 1.26.3 | deploy | GCP | Fast | 4.6.0
7 | OOTB 1.25 cluster | docker | baseline,viya,install | front door TLS | static kubeconfig | n/a | 428.0.0 | 0.5.2 | 1.25.8 | 1.25.8-gke | deploy | GCP | stable:2023.04 | 4.6.0
8 | OOTB 1.24 cluster | docker | baseline,viya,install | front door TLS | static kubeconfig | n/a | 428.0.0 | 0.5.2 | 1.25.8 | 1.24.12-gke.1000 | DO | GCP | stable:2023.04 | 4.6.0

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ RUN curl -sLO https://storage.googleapis.com/kubernetes-release/release/v{$kubec
 FROM baseline
 ARG helm_version=3.9.4
 ARG aws_cli_version=2.7.22
-ARG gcp_cli_version=427.0.0-0
+ARG gcp_cli_version=428.0.0-0
 
 # Add extra packages
 RUN apt-get install -y gzip wget git git-lfs jq sshpass skopeo rsync \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,13 @@
 # syntax=docker/dockerfile:experimental
 FROM ubuntu:22.04 as baseline
-RUN apt update && apt upgrade -y \
-  && apt install -y python3 python3-dev python3-pip curl unzip \
+
+RUN apt-get update && apt-get upgrade -y \
+  && apt-get install -y python3 python3-dev python3-pip curl unzip apt-transport-https ca-certificates gnupg \
   && update-alternatives --install /usr/bin/python python /usr/bin/python3 1 \
   && update-alternatives --install /usr/bin/pip pip /usr/bin/pip3 1
 
 FROM baseline as tool_builder
-ARG kubectl_version=1.24.10
+ARG kubectl_version=1.25.8
 
 WORKDIR /build
 
@@ -16,10 +17,10 @@ RUN curl -sLO https://storage.googleapis.com/kubernetes-release/release/v{$kubec
 FROM baseline
 ARG helm_version=3.9.4
 ARG aws_cli_version=2.7.22
-ARG gcp_cli_version=409.0.0
+ARG gcp_cli_version=427.0.0-0
 
 # Add extra packages
-RUN apt install -y gzip wget git git-lfs jq sshpass skopeo rsync \
+RUN apt-get install -y gzip wget git git-lfs jq sshpass skopeo rsync \
   && curl -ksLO https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 && chmod 755 get-helm-3 \
   && ./get-helm-3 --version v$helm_version --no-sudo \
   && helm plugin install https://github.com/databus23/helm-diff \
@@ -30,9 +31,10 @@ RUN apt install -y gzip wget git git-lfs jq sshpass skopeo rsync \
   # AZURE
   && curl -sL https://aka.ms/InstallAzureCLIDeb | bash \
   # GCP
-  && curl "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-${gcp_cli_version}-linux-x86_64.tar.gz" -o gcpcli.tar.gz \
-  && tar -xvf gcpcli.tar.gz \
-  && ./google-cloud-sdk/install.sh
+  && echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list \
+  && curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add - \
+  && apt-get update && apt-get install google-cloud-cli:amd64=${gcp_cli_version} \
+  && apt-get install google-cloud-sdk-gke-gcloud-auth-plugin 
 
 COPY --from=tool_builder /build/kubectl /usr/local/bin/kubectl
 
@@ -50,7 +52,6 @@ RUN pip install -r ./requirements.txt \
 ENV PLAYBOOK=playbook.yaml
 ENV VIYA4_DEPLOYMENT_TOOLING=docker
 ENV ANSIBLE_CONFIG=/viya4-deployment/ansible.cfg
-ENV PATH=$PATH:/google-cloud-sdk/bin/
 
 VOLUME ["/data", "/config", "/vault"]
 ENTRYPOINT ["/viya4-deployment/docker-entrypoint.sh"]
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -23,5 +23,7 @@ do
   fi
 done
 
+# TODO: Can remove the next line when DAC moves to kubectl >= 1.26
+export USE_GKE_GCLOUD_AUTH_PLUGIN=True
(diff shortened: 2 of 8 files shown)


docs/user/Dependencies.md
```diff
@@ -24,6 +24,12 @@ The following list details our dependencies and versions (~ indicates multiple p
 | ansible-galaxy | ansible.utils    | 2.3.0       |
 | ansible-galaxy | kubernetes.core  | 2.3.2       |
 
+If you are using a provider based kubeconfig file created by viya4-iac-gcp:4.5.0 or newer, install these dependencies:
+| SOURCE         | NAME                    | VERSION     |
+|----------------|-------------------------|-------------|
+| ~              | gcloud                  | 428.0.0     |
+| ~              | gcloud-gke-auth-plugin  | 428.0.0     |
```
jarpat, line 31: I think this should be 0.5.2 or >

jarpat (changes requested):

docs/user/Dependencies.md
dhoucgitter: @jarpat, thanks, I updated the table.

riragh (approved):

Dockerfile
```diff
@@ -1,12 +1,13 @@
 # syntax=docker/dockerfile:experimental
 FROM ubuntu:22.04 as baseline
-RUN apt update && apt upgrade -y \
-  && apt install -y python3 python3-dev python3-pip curl unzip \
+
+RUN apt-get update && apt-get upgrade -y \
+  && apt-get install -y python3 python3-dev python3-pip curl unzip apt-transport-https ca-certificates gnupg \
```
thpang, line 5: I believe we currently use `apt` instead of `apt-get` in the building of our docker images. I am not opposed to this, but again who tracking to keep this consistent across all our repos if we change here?

jarpat (approved):

docs/user/Kubeconfig.md
```diff
@@ -0,0 +1,52 @@
+# Kubernetes Configuration File Types
+
+## Overview
+
+### Notes - viya4-deployment:6.6.0
```
thpang, line 5: Are we now keeping track of versioned information in these markdown files? This is going to add quite a bit of support and update for each deployment. Again, not opposed just pointing out that over the long-term this fill will get pretty big capturing version info.

docker-entrypoint.sh
```diff
@@ -23,5 +23,7 @@ do
   fi
 done
 
+# TODO: Can remove the next line when DAC moves to kubectl >= 1.26
```
thpang, line 26: This release is to support kubernetes 1.26. I would suggest saying that this will be removed when the default value is moved to 1.26 and greater.

docs/user/Kubeconfig.md
```diff
@@ -0,0 +1,52 @@
...
+
+The release of kubectl v1.26 is dropping support for built-in provider-specific code in their project for authentication and instead opting for a plugin-based strategy. A quote from the [Google blog](https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke) provides an explanation for this change:
+
+>To ensure the separation between the open source version of Kubernetes and those versions that are customized by services providers like Google, the open source community is requiring that all provider-specific code that currently exists in the OSS code base be removed starting with v1.26.
+
+### Usage with provider based kubeconfig files from viya4-iac-gcp:4.5.0
```
thpang, line 11: Again version info concern.

Dockerfile
dhoucgitter: @thpang, I did a quick scan and found that iac-gcp and iac-k8s are the two other projects using apt, gcp uses apt-get and iac-k8s uses both apt and apt-get. aws is using yum and azure is using apk based on their base image types.

The change to use apt-get was based on this hadolint recommendation:

./Dockerfile.trial:23 DL3027 warning: Do not use apt as it is meant to be a end-user tool, use apt-get or apt-cache instead

How about creating a backlog ticket to update iac-k8s so all three are in sync?

docs/user/Kubeconfig.md
dhoucgitter: I noticed that we used versions in existing doc already like https://github.com/sassoftware/viya4-iac-gcp/pull/169/files#diff-8a22ddc8bcdf8b5b392d4e76e637e4e6fd79cba9ebf343dddef54713b947a293
It does seem helpful to include the version here since our doc is meant to provide relevant info for current and past versions. 
In this case the version is meant to imply 6.6.0 or later (although not stated explicity) so it would not necessarily have to be updated each release, if we included that qualifier does that make it any less of an issue?

docs/user/Kubeconfig.md
dhoucgitter: See reply above https://github.com/sassoftware/viya4-deployment/pull/435#discussion_r1187861765 for rationale and options.

docker-entrypoint.sh
dhoucgitter: I applied your suggestion.

docs/user/Kubeconfig.md
thpang: Alright. I am good with that.

thpang (approved):
Thanks for the clarifications

PR #320, opened 2022-10-20
5.6.0 - October 20, 2022

riragh:
- #303 
- #307 
- #311
- #312
- #313
- #318
- #319

PR #446, opened 2023-05-19
feat: (IAC-1096) add support for Azure service bus

riragh:
# Changes:
Support added for  Message broker: `Azure service bus` to be used by SAS Viya Platform deployment instead of RabbitMQ.
The changes to create Azure service bus were added in `viya4-iac-azure`: [PR #320](https://github.com/sassoftware/viya4-iac-azure/pull/320)

## Variables introduced in this change:
**Note:** This is currently Experimental/Undocumented feature this will announced along with SAS Viya feature availability.

| V4_CFG_MESSAGE_BROKER_ENABLE| Enables the message broker created in your cloud provider. Default is set to false which uses RabbitMQ| string |false | false |Currently only supported for Azure's message broker: `Azure Service Bus` | viya |
| V4_CFG_MESSAGE_BROKER_HOST | Hostname to access the message broker| string | | false | | viya |
| V4_CFG_MESSAGE_BROKER_NAME | Specifies the message broker's name| string | | false | | viya |
| V4_CFG_MESSAGE_BROKER_PASSWORD | The message broker's primary key or password| string | | false | | viya |

# Tests:
Verified following on the scenarios listed below, please see internal ticket for details.
- Verified cluster creation was successful.
- Azure Service bus was created. 
- SAS Viya Platform consumed the Azure service bus information from state file correctly. 
- Rabbitmq pods were not created.

**Note:** SAS Viya Platform services are still not ready. Almost all pods stabilized except for few services. Launcher service is failing with `[BROKER_MISSING_USERNAME]` causing other services to fail. This is expected until all the services are updated to use the new Azure Service bus instead of Rabbitmq.

|Scenario|Task|Deploy method|Cadence|
|:----|:----|:----|:----|
|1|Defaults, `V4_CFG_MESSAGE_BROKER_ENABLE= true`, internal Postgres |ansible|fast:2020|
|2|`V4_CFG_MESSAGE_BROKER_ENABLE= true`, external Postgres|ansible|fast:2020|
|3|`V4_CFG_MESSAGE_BROKER_ENABLE= true`. Values used from tfstate: `message_broker_name= "ritika1"`, `aks_cluster_sku_tier = "Standard"`, external Postgres|ansible|fast:2020|


thpang (approved):
Same comment as with the other changes. This should have been more generic and cloud agnostic or at least investigated and reasoning why the naming was kept azure specific especially since we'll probably need to handle this on AWS/GCP in the future.

roles/vdm/tasks/main.yaml
```diff
@@ -193,6 +193,14 @@
     - uninstall
     - update
 
+- name: Include Message broker
+  include_tasks: message_broker.yaml
+  when: PROVIDER == "azure"
```
jarpat, line 198: Should an additional check be added here to make sure `V4_CFG_MESSAGE_BROKER_HOST`, `V4_CFG_MESSAGE_BROKER_NAME`, & `V4_CFG_MESSAGE_BROKER_PASSWORD` is not null, otherwise I think the task in L13 in roles/vdm/tasks/message_broker.yaml would update the sitedefault.yaml with keys that have null values?

riragh: Good point. I will update the task with additional conditions here: https://github.com/sassoftware/viya4-deployment/blob/IAC-1020/roles/vdm/tasks/message_broker.yaml#L25

riragh: @jarpat Updated the code to add additional guards to add values to sitedefault only if not none. Also introduced another variable `V4_CFG_MESSAGE_BROKER_ENABLE` which will allow user to choose whether want the Cloud message broker like Azure service bus or continue with the default RabbitMQ.

riragh:
> Same comment as with the other changes. This should have been more generic and cloud agnostic or at least investigated and reasoning why the naming was kept azure specific especially since we'll probably need to handle this on AWS/GCP in the future.

Same reply as with the other repo. After investigation across cloud for message brokers, these are available options:
```
Azure: Azure Service Bus
AWS: Amazon MQ
GCP: Google Pub/Sub
```
To keep variables and module names generic I went with name as `message broker`. Azure service bus provides option to create queues or topics and SAS Viya Platform creates `Topics`. 
I thought it would be confusing to name the variables/modules as `message queue` when we don't create queues.
I don't mind changing variable/module names, please let me know.

dhoucgitter (approved):

jarpat (approved):

PR #448, opened 2023-05-24
feat: (IAC-1014) Add Ability to Generate the site.yaml file

jarpat:
### Changes

**Experimental:** Adds the ability to generate the site.yaml for kubernetes commands based deployments.

To generate the site.yaml, you must first the opt to use the `sas-orchestration deploy` deployment method by setting our existing variable, `V4_DEPLOYMENT_OPERATOR_ENABLED:false` in your ansible-vars.yaml. Additionally, you must set the `V4_SAS_ORCHESTRATION_USE_SITE_YAML:true` variable in your ansible-vars.yaml to enable the site.yaml generation.

NOTE: At this point in time we are **ONLY** generating the site.yaml file so a user could potentially perform a deployment using the "kubernetes commands" method. Changes are planned to make use of the site.yaml later.

### Tests

| Scenario | Provider | K8s Version     | Deployment Method | Order  | Cadence   | DEPLOY | V4_DEPLOYMENT_OPERATOR_ENABLED | V4_SAS_ORCHESTRATION_USE_SITE_YAML | MT Tasks            | Notes                                                                                                                                     |
|----------|----------|-----------------|-------------------|--------|-----------|--------|--------------------------------|------------------------------------|---------------------|-------------------------------------------------------------------------------------------------------------------------------------------|
| 1        | GCP      | v1.26.4-gke.500 | Docker            | * | fast:2020 | true   | false                          | true                               | n/a                 | site.yaml generated                                                                                                                       |
| 2        | GCP      | v1.26.4-gke.500 | Ansible           | * | fast:2020 | true   | false                          | true                               | n/a                 | site.yaml generated                                                                                                                       |
| 3        | GCP      | v1.26.4-gke.500 | Docker            | * | fast:2020 | true   | false                          | false                              | n/a                 | site.yaml not generated since V4_SAS_ORCHESTRATION_USE_SITE_YAML:false                                                                    |
| 4        | GCP      | v1.26.4-gke.500 | Docker            | * | fast:2020 | true   | true                           | true                               | n/a                 | site.yaml not generated since user opted to use DO with V4_DEPLOYMENT_OPERATOR_ENABLED:true                                                                   |
| 5        | GCP      | v1.26.4-gke.500 | Ansible            | * | fast:2020 | true   | false                          | true                               | onboard,cas-onboard | site.yaml regenerated after tenant resources added to kustomization.yaml                                                                  |
| 6        | GCP      | v1.26.4-gke.500 | Docker            | * | fast:2020 | false  | false                          | false                              | n/a                 | Nothing was deployed into the cluster, mimics the workflow a user would take if they wanted to perform the kubernetes commands themselves |

### DIFF
--- a/roles/vdm/defaults/main.yaml
+++ b/roles/vdm/defaults/main.yaml
@@ -101,7 +101,7 @@ V4_DEPLOYMENT_OPERATOR_CRB: sasoperator
 
 # SAS Orchestration Command
 # Experimental
-V4_SAS_ORCHESTRATION_USE_SITE_YAML: false
+V4_SAS_ORCHESTRATION_USE_SITE_YAML: true
 
 ## Below the line deployment -- internal use only
 ## Setting true enables using custom du for below the line testing
--- a/roles/vdm/defaults/main.yaml
+++ b/roles/vdm/defaults/main.yaml
@@ -99,6 +99,9 @@ V4_DEPLOYMENT_OPERATOR_SCOPE: cluster
 V4_DEPLOYMENT_OPERATOR_NAMESPACE: sasoperator
 V4_DEPLOYMENT_OPERATOR_CRB: sasoperator
 
+# SAS Orchestration Command
+V4_SAS_ORCHESTRATION_USE_SITE_YAML: false
+
 ## Below the line deployment -- internal use only
 ## Setting true enables using custom du for below the line testing
 V4_CFG_BELOW_THE_LINE: false
--- a/roles/vdm/defaults/main.yaml
+++ b/roles/vdm/defaults/main.yaml
@@ -100,6 +100,7 @@ V4_DEPLOYMENT_OPERATOR_NAMESPACE: sasoperator
 V4_DEPLOYMENT_OPERATOR_CRB: sasoperator
 
 # SAS Orchestration Command
+# Experimental
 V4_SAS_ORCHESTRATION_USE_SITE_YAML: false
 
 ## Below the line deployment -- internal use only
--- a/roles/vdm/defaults/main.yaml
+++ b/roles/vdm/defaults/main.yaml
@@ -101,7 +101,7 @@ V4_DEPLOYMENT_OPERATOR_CRB: sasoperator
 
 # SAS Orchestration Command
 # Experimental
-V4_SAS_ORCHESTRATION_USE_SITE_YAML: true
+V4_SAS_ORCHESTRATION_USE_SITE_YAML: false
 
 ## Below the line deployment -- internal use only
 ## Setting true enables using custom du for below the line testing
--- a/roles/vdm/tasks/main.yaml
+++ b/roles/vdm/tasks/main.yaml
@@ -223,6 +223,19 @@
     - update
     - multi-tenancy
 
+# Only applicable for sas-orchestration based deployments
+# when the V4_SAS_ORCHESTRATION_USE_SITE_YAML flag is set
+- name: Include SAS Viya Kubernetes Manifest
+  include_tasks: sas_viya_kubernetes_manifest.yaml
+  when:
+    - not V4_DEPLOYMENT_OPERATOR_ENABLED
+    - V4_SAS_ORCHESTRATION_USE_SITE_YAML
+  tags:
+    - install
+    - uninstall
+    - update
+    - multi-tenancy
+
 - name: Include Deploy
   include_tasks: deploy.yaml
   when:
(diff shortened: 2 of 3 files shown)


jarpat:
Same tests table as above, putting a markdown table in the comments vs the PR description doesn't have a word wrap issue and uses a scroll bar instead.



| Scenario | Provider | K8s Version     | Deployment Method | Order  | Cadence   | DEPLOY | V4_DEPLOYMENT_OPERATOR_ENABLED | V4_SAS_ORCHESTRATION_USE_SITE_YAML | MT Tasks            | Notes                                                                                                                                     |
|----------|----------|-----------------|-------------------|--------|-----------|--------|--------------------------------|------------------------------------|---------------------|-------------------------------------------------------------------------------------------------------------------------------------------|
| 1        | GCP      | v1.26.4-gke.500 | Docker            | 0B2F9L | fast:2020 | true   | false                          | true                               | n/a                 | site.yaml generated                                                                                                                       |
| 2        | GCP      | v1.26.4-gke.500 | Ansible           | 0B2F9L | fast:2020 | true   | false                          | true                               | n/a                 | site.yaml generated                                                                                                                       |
| 3        | GCP      | v1.26.4-gke.500 | Docker            | 0B2F9L | fast:2020 | true   | false                          | false                              | n/a                 | site.yaml not generated since V4_SAS_ORCHESTRATION_USE_SITE_YAML:false                                                                    |
| 4        | GCP      | v1.26.4-gke.500 | Docker            | 0B2F9L | fast:2020 | true   | true                           | true                               | n/a                 | site.yaml not generated since user opted to use DO with V4_DEPLOYMENT_OPERATOR_ENABLED:true                                                                   |
| 5        | GCP      | v1.26.4-gke.500 | Ansible            | 0B2F9L | fast:2020 | true   | false                          | true                               | onboard,cas-onboard | site.yaml regenerated after tenant resources added to kustomization.yaml                                                                  |
| 6        | GCP      | v1.26.4-gke.500 | Docker            | 0B2F9L | fast:2020 | false  | false                          | false                              | n/a                 | Nothing was deployed into the cluster, mimics the workflow a user would take if they wanted to perform the kubernetes commands themselves |

roles/vdm/defaults/main.yaml
```diff
@@ -99,6 +99,10 @@ V4_DEPLOYMENT_OPERATOR_SCOPE: cluster
 V4_DEPLOYMENT_OPERATOR_NAMESPACE: sasoperator
 V4_DEPLOYMENT_OPERATOR_CRB: sasoperator
 
+# SAS Orchestration Command
+# Experimental
+V4_SAS_ORCHESTRATION_USE_SITE_YAML: false
```
thpang, line 104: For the use cases on this one. Do we have other parameters that are used for the Orchestration CLI or is this the only option we surface to customers? Why false. The goal here is to help folks move forward. By having set to false, the current customer support teams will not see any difference if they don't include this item. If we're supporting site.yaml it should always be there. The other option is to have the orchestration command line generated the sasdeployment.yaml file for use, but from all of the conversations around this one, I believe it's not needed if we're using site.yaml. Thoughts?

jarpat: Had a offline discussion with @thpang, updated this flag to "true"

So the updated behavior would be:

- Using the deployment operator as the deployment option (`V4_DEPLOYMENT_OPERATOR_ENABLED:true`) the site.yaml is still never generated. No change here.

- Using the sas-orchestration deploy command as the deployment option (`V4_DEPLOYMENT_OPERATOR_ENABLED:false`) the site.yaml will be generated and used (in the future) by default. The user **will** have to to turn this option off (`V4_SAS_ORCHESTRATION_USE_SITE_YAML:false`) if they instead want to use CRD with the sas-orchestration deploy command.

Not marking this as resolved, conversation planned with with the orchestration team for further clarifcation.

jarpat: Changing this back to false, we are **ONLY** generating the site.yaml at this point in time and not using it with the sas-orchestration deploy command. 

Since the file cannot actually be used by viya4-deployment and fed into the sas-orchestration deploy command until those changes are in place upstream, it makes sense to leave this false for now.

When we are able to use the site.yaml file with the sas-orchestration deploy we can considering toggling this true by default.

@thpang let me know if you agree.

jarpat:
Closing this since I'm reworking the code.
This is an experimental orchestration feature so it will exist in a dev branch and will be released once fully internally tested.

PR #461, opened 2023-07-19
fix: (IAC-1004) Remove sas-connect-customer-provided-ingress-certificate-tls-certframe-transformer.yaml

dhoucgitter:
# Changes
Documentation from a past internal ticket indicates that the subject transformer yaml file was removed from the product and is not being delivered. Comments indicate that transformer was originally designed to replace the connect certificate with a customer provided certificate, although that is not a currently supported option. DAC should follow suit and remove its template for that transformer to avoid kustomize build problems that have been identified when including it.

- Preserve the addition of V4_CFG_CONNNECT_FQDN to V4_CFG_TLS_ADDITIONAL_SAN_DNS variable
- Remove addition of sas-connect-customer-provided-ingress-certificate-tls-certframe-transformer.yaml transformer to the kustomization.yaml per the explanation above.

# Tests

|Scenario|Task|method|Provider|k8s version|Cadence|tasks|Deploy method|Remarks|
|--|--|--|--|--|--|--|--|--|
|1|deployment |Ansible|Azure|1.26.3|stable:2023.06|baseline,viya,install|deploy command|Deployment completed successfully with all pods stabilized. Successful viya_admin login to Viya SASDrive application.|
|2|kustomize build . -o site.yaml| See remarks below |n/a |n/a |n/a |n/a |n/a | Successful|

Post deployment, navigate to the \<$BASE_DIR\>/\<cluster\>/\<namespace\> folder containing the DAC generated kustomization.yaml file and verified that `kustomize build . -o site.yaml` runs (without encountering the error described in the internal ticket) and successfully generates the site.yaml manifest file.

Re-tested both scenarios using 60acb1f commit with successful results for both.


jarpat (approved):

roles/vdm/tasks/connect.yaml
```diff
@@ -20,13 +20,6 @@
   - name: connect - SAS/CONNECT FQDN to the SAN DNS list for cert-manager
     set_fact:
       V4_CFG_TLS_ADDITIONAL_SAN_DNS: "{{ V4_CFG_TLS_ADDITIONAL_SAN_DNS }} {{ V4_CFG_CONNECT_FQDN }}"
```
thpang, line 22: Is the block still needed? Also, is the content needed here for the SAS/CONNECT FQDN? Just wondering if we don't have the transformer does this code help if the customer add this on their own.

thpang (changes requested):

roles/vdm/tasks/connect.yaml
dhoucgitter: I can check into that further, I had initially removed it but then restored it to preserve the behavior since it was contributing to the manifest creation failure. Suspect we will want to update the CONFIG-VARS.md documentation surrounding that variable if we end up removing that block.

anthoday (approved):

roles/vdm/tasks/connect.yaml
dhoucgitter: @thpang, I removed the block per our discussion on Weds morning and restored the condition to what it was before the block that I removed was added.

dhoucgitter: Resolved the merge conflicts by rebasing these changes to origin/staging just now

thpang (approved):

PR #488, opened 2023-09-13
docs: (IAC-1132) Add License Update Note

jarpat:
### Changes

Add note for users who will be updating their Software License, they need to make a modification to their kustomization.yaml


dhoucgitter (approved):

riragh (approved):

jarpat:
After some external conversation with @thpang this doc may not be required since it's mixing and matching DAC & manual deployments which is out of scope. 
DAC may be able to update the license itself, will need to do something testing.

jarpat:
Closing. See Internal Ticket.

PR #498, opened 2023-10-17
docs: (IAC-1202) PostgreSQL Transfer Tooling Supplemental Doc

dhoucgitter:
### Changes

- Move supplemental doc for users who followed the PostgreSQL Data Transfer Guide manual steps to migrate to an external Postgres Cluster and want to their deployment managed by viya4-deployment again to a separate .md file in the docs/user folder.
- Add links to all three cloud provider PostgreSQL database server configuration sections.
- Update steps to recommend modifying the V4_CFG_POSTGRES_SERVERS map `default:` entry to reference an external database server by changing it to `internal: false` instead of completely removing the server map entry.


sayeun:
May also be helpful to add a link to the DAC Postgres section.

dhoucgitter:
> May also be helpful to add a link to the DAC Postgres section.

@sayeun, there is a link to the DAC PostgreSQL CONFIG-VARS.md section [here](https://github.com/sassoftware/viya4-deployment/pull/498/files#diff-2fba30c62c436a801c23021030f10f35f0b6ead601db5ab0fc28c572e81e6035R28) in the context of provisioning an external PostgreSQL cluster without using Viya4 IAC projects. Would you suggest adding a DAC PostgreSQL link along with the IAC PostgreSQL links, or elsewhere?

jkmin96:
About "### Using IAC to create an external PostgreSQL database server"
Those PG instructions are based on the OOTB.
Can user just run IAC to add the external PG w/o doing any extra work for Viya? As I mentioned yesterday, I thought that stopping Viya is a requirement before updating the existing cluster. Please make sure if stopping Viya is not necessary.

dhoucgitter:
> About title "## PostgreSQL Data Transfer Tool for the SAS Viya Platform - viya4-deployment Post-Transfer Steps (Experimental)" I'm still not sure it would be o.k. to use "Post-transfer" term for this doc.
> 
> About "### Using IAC to create an external PostgreSQL database server" Those PG instructions are based on the OOTB. Can user just run IAC to add the external PG w/o doing any extra work for Viya? As I mentioned yesterday, I thought that stopping Viya is a requirement before updating the existing cluster. Please make sure if stopping Viya is not necessary.

If the expected terraform plan/apply changes that are listed after only adding the PostgreSQL server map to their terraform-input.tfvars include destroying or re-creating a part of the cluster that Viya is currently using, for example substantive changes to existing resources other than the new PostgreSQL database server and recreating the jump server, then stopping Viya first makes sense. If the expected changes are just creating the PostgreSQL database server and jump server, I would not expect stopping Viya to be necessary. Since that advice can be left to some interpretation, perhaps we should always recommend stopping Viya, creating the new PostgreSQL server, then restarting Viya as the way to proceed. Thoughts?

jkmin96 (approved):
Please check if title is o.k. to include "post" since this doc includes about adding external PG.

Please check if stopping Viya is required when user adds external PGs to an existing cluster using IAC. If not, current doc is o.k.

sayeun (approved):
LGTM

riragh (approved):

thpang (approved):

PR #530, opened 2024-03-05
feat: (IAC-1236) Dependency version update

riragh:
### Changes
Updates 3rd party dependencies. Consumers of the Dockerfile will automatically have these updated dependencies installed, and users who directly run this project on this host will need to update the dependencies themselves.

#### Binaries
* `kubectl` 1.27.9 -> 1.27.11
* `helm` 3.13.2 -> 3.14.2
* `aws_cli_version` 2.13.33 -> 2.15.22
* `gcp_cli_version` 460.0.0 -> 464.0.0

#### Python Dependencies 
* `ansible`  9.1.0 -> 9.2.0
  * This change updates the `ansible-core` version to 2.16.4
* `kubernetes` 26.1.0 -> 27.2.0
  * Not a breaking change. Used by `kubernetes.core`, updated to match the kubernetes version `1.27`.
  * https://github.com/kubernetes-client/python?tab=readme-ov-file#compatibility
* `dnspython` 2.3.0 -> 2.6.1
* `openshift` 0.13.1 -> 0.13.2
* `docker` 5.0.3 -> 7.0.0
  * Major version update, there were few deprecations, verified through testing those do not affect us.

#### Ansible Dependencies
* `ansible.utils` 2.6.1 -> 3.1.0
  * Major version update: the breaking change in this collection is requires_ansible to >=2.14.0, which we already meet so does not affect us.
* `community.docker` 2.7.8 -> 3.8.0
  * Major version update, however we also install the Docker SDK for Python, the requirements are met and does not cause any breaking change for us
* `kubernetes.core` 2.3.2 -> 2.4.0
  * Major version update was to drop support for ansible-core versions lower than 2.14, Python versions lower than 3.9 and updates python kubernetes library to 24.2.0, helm/kind-action to 1.8.0, kubernetes >= 1.24, along with bug fixes and minor changes. As we are already using ansible-core > 2.14 this change doesn't affect us.

### Tests

| Scenario | Provider | Kubectl version |  Order  | Cadence   | Notes   |
|----------|----------|-------------|-----------------|-----------|--------------------------|
| 1        | Azure      | 1.27.11     | ****** | fast:2020 | OOTB    |
| 2        | Azure      | 1.27.11      | ****** | fast:2020 | Downgraded project deps to 6.17.0 to verify compatibility  |
| 3        | Azure      | 1.27.11      | ****** | fast:2020 |   OOTB     |


thpang (approved):
LGTM

dhoucgitter (approved):
LGTM

Only suggestion would be to also mention the version of ansible-core that moving to ansible 9.2.0 brings us to. I see that you already mentioned that ansible-core >= 2.14 is required by some of the new deps but also saying which version of ansible-core we are now at would be useful too.

riragh:
> LGTM
> 
> Only suggestion would be to also mention the version of ansible-core that moving to ansible 9.2.0 brings us to. I see that you already mentioned that ansible-core >= 2.14 is required by some of the new deps but also saying which version of ansible-core we are now at would be useful too.

Updated the description to indicate that ansible-core version gets updated to 2.16.4.

jarpat (approved):

PR #532, opened 2024-03-11
docs: (IAC-1294) Add Note about sitedefault When Not Using Embedded LDAP

jarpat:
### Changes
 Doc update about users needing to set `V4_CFG_SITEDEFAULT` if they opt not to use the embedded LDAP server.

### DIFF
--- a/docs/CONFIG-VARS.md
+++ b/docs/CONFIG-VARS.md
@@ -351,7 +351,7 @@ Additional documentation for the SAS Workload Orchestrator Service can be found 
 | Name | Description | Type | Default | Required | Notes | Tasks |
 | :--- | ---: | ---: | ---: | ---: | ---: | ---: |
 | V4_CFG_CLUSTER_NODE_POOL_MODE | The mode of cluster node pool to use | string | "standard" | false | [standard, minimal] | viya |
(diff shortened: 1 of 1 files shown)


thpang (approved):
LGTM

docs/CONFIG-VARS.md
```diff
@@ -351,7 +351,7 @@ Additional documentation for the SAS Workload Orchestrator Service can be found
 | Name | Description | Type | Default | Required | Notes | Tasks |
 | :--- | ---: | ---: | ---: | ---: | ---: | ---: |
 | V4_CFG_CLUSTER_NODE_POOL_MODE | The mode of cluster node pool to use | string | "standard" | false | [standard, minimal] | viya |
-| V4_CFG_EMBEDDED_LDAP_ENABLE | Deploy OpenLDAP in the namespace for authentication | bool | false | false | [Openldap Config](../roles/vdm/templates/generators/openldap-bootstrap-config.yaml) | viya |
+| V4_CFG_EMBEDDED_LDAP_ENABLE | Deploy OpenLDAP in the namespace for authentication | bool | false | false | [Openldap Config](../roles/vdm/templates/generators/openldap-bootstrap-config.yaml). If you set this value to false, you must set `V4_CFG_SITEDEFAULT` to point to a sitedefault file which contains values applicable for your authentication configuration. | viya |
```
dhoucgitter, line 354: Since the default for `V4_CFG_EMBEDDED_LDAP_ENABLE` is false, should we alter the wording to say it a bit differently, like "If you do not set this value to true, you must set `V4_CFG_SITEDEFAULT` to point to a ..."

jarpat: I agree, that makes sense since the default value will be false, so a user may not explictly "set" this value. 

Updated the wording

dhoucgitter (approved):
LGTM

PR #552, opened 2024-05-20
docs: (IAC-1462) remove references to multi-tenancy documentation

riragh:
## Changes:
SAS Viya currently has pause the support for Multi-tenancy, so all the references to Multi-tenancy documentation have been removed.


jarpat:
@riragh should the example file be removed as well?

thpang (approved):

dhoucgitter:
> @riragh should the example file be removed as well?

good point, not sure why we'd want to keep that if deleting the doc files.

riragh:
So the idea is to keep the code in as is, just remove the doc references. So this would be like undocumented feature. I am not sure if example file would be considered as doc or code. If we you all think we should remove I am good to remove. Please let me know.

dhoucgitter:
> @jarpat, @dhoucgitter So the idea is to keep the code in as is, just remove the doc references. So this would be like undocumented feature. I am not sure if example file would be considered as doc or code. If we you all think we should remove I am good to remove. Please let me know.

I suspect that if our intent is that the MT behavior/code should still be in place to support using Multi-Tenancy, just not document how to do that, then removing the example file is probably not a good idea.

dhoucgitter (approved):

jarpat (approved):

PR #558, opened 2024-06-17
feat: (IAC-1476) DAC - Security scan 2024.06

dhoucgitter:
Update tools and cli version to remediate Critical and many High security vulnerabilities reported by the recent Aqua Scan, see full results in the internal ticket.

Final scan result has: 0 Critical, 18 High security vulnerabilities.


thpang (approved):

Dockerfile
```diff
@@ -54,7 +54,7 @@ RUN pip install -r ./requirements.txt \
   && pip cache purge \
   && chmod -R g=u /etc/passwd /etc/group /viya4-deployment/ \
   && chmod 755 /viya4-deployment/docker-entrypoint.sh \
-  && git config --system --add safe.directory /viya4-deployment
+  && git config --system --add safe.directory /viya4-deployment ||:
```
jarpat, line 57: Was the addition`||:` needed? Shouldn't we stop the Docker build if this command fails rather than ignoring the error?

dhoucgitter: @jarpat, it is a change that I often make to allow building the docker image if the folder docker build is run from is not a first-class repository folder, ie. a work-tree folder. If that is the case, the `git config --system` command will fail but and the docker image will fail to create, however, since this line was added as a workaround to fix a git cli vulnerability in an earlier git version, it may be OK to completely remove this line now. I will do a quick search and post back, interested on your thoughts with doing that and what risk we might incur if any. I have run the resulting `viy4-deployment` container with that change in the past with no obvious ill effect.

jarpat: I think that change is fine now that I'm aware of the reasoning behind it.

jarpat (approved):

PR #572, opened 2024-09-12
Enhancements and Basic OCP Support

NormJohnIV:
1. Removed the need to set `vdm: true` when calling overlay_facts
2. Added new vdm helper `add_if_exists` that can add overlays if a defined file exists
3. Granted users more control over ordering their overlays via prefixing files with two digit number. The kustomization.yaml output has also been updated to display the priorites of all overlays and their source (either `vdm` or `user`)
ex. 57-my-example-overlay.yaml
4. Add basic support for Openshift. This includes routes, tls, remove seccomp profile, and optionally SecurityContextConstraints with the dependent RBAC permissions.


saschjmil:
Thanks for the contribution @NormJohnIV. I'm opening a PM feature request for the OpenShift support. While we wait for that, would you be willing to move the enhancements into a new PR? That would help us merge them into the main branch quicker.

saschjmil:
OpenShift support is currently on hold. I'll keep the PM Feature Request label on this PR while we wait. Meanwhile, you are welcome to move the enhancements into another PR.

PR #586, opened 2024-12-12
feat: (PSKD-957) ingress-nginx configmap changes for v1.12+

dhoucgitter:
### Changes
- Add two config map values to maintain the default values they had prior to `ingress-nginx` v1.12
- `ingress-nginx` v1.12.0 has changed the defaults for those two config map values, the values required by the Viya platform differ from the new defaults.

### Tests
To be performed when `ingress-nginx` v1.12.0 is released.


saschjmil (approved):
LGTM

ndickens (approved):

ephesused (approved):

PR #594, opened 2025-02-18
feat: Use block storage for Crunchy Postgres and RabbitMQ for AWS (PSKD-1282)

dhoucgitter:
## Changes
- For AWS, create block storage classes for RabbitMQ and Crunchy Postgres use so that persistent volume claim requests in those applications create elastic block storage based persistent volumes instead of using NFS based persistent volumes for those applications.
- For AWS, configure the two new io2 based storage classes with reasonable iopsPerGB and Throughput defaults based on volume sizes for satisfactory RabbitMQ and Crunchy Postgres performance. Allow for configuration variable overrides to choose different volume types or performance related parameters for the new storage classes if needed.


saschjmil (approved):
Looks good!

docs/CONFIG-VARS.md
```diff
@@ -411,6 +416,16 @@ The EBS CSI driver is currently only used for kubernetes v1.23 or later AWS EKS
...
+|EBS_CSI_CRUNCHY_STORAGE_CLASS_NAME| The EBS CSI storage class name for Crunchy Postgres use| string| io2-vol-pg | false | | baseline |
+|EBS_CSI_CRUNCHY_STORAGE_CLASS_VOLUME_TYPE| The EBS CSI volume type to use for Crunchy Postgres persistent volumes | string | io2 | false | Supported values: [`io2`, `io1`, `gp3`] | baseline |
+|EBS_CSI_CRUNCHY_STORAGE_CLASS_IOPSPERGB | IOPs per GB parameter for the `EBS_CSI_CRUNCHY_STORAGE_CLASS_NAME` storage class | string | 40 | false |Multiply this value by the volume size in GiB to obtain total IOPS per volume | baseline |
+|EBS_CSI_CRUNCHY_STORAGE_CLASS_THROUGHPUT | Maximum volume throughput in MiB/s for the `EBS_CSI_CRUNCHY_STORAGE_CLASS_NAME` storage class | string| 400 | false | The maximum value for io2, io1 and gp3 volume types is 1000.| baseline |
+|EBS_CSI_CRUNCHY_STORAGE_CLASS_RECLAIM_POLICY | The ReclaimPolicy for the `EBS_CSI_CRUNCHY_STORAGE_CLASS_NAME` storage class | string | Delete | false | Supported values: [`Delete`, `Retain`] | baseline |
```
joboon, line 427: There's not a similar reclaim policy configuration for rabbit. Just curious: was that by design?

joboon (approved):
Looks good to me, just a question for my own understanding.

docs/CONFIG-VARS.md
joboon: Understood. It feels like the option to do so on Rabbit would have no cost, but I understand that it has historically had no audience, and thus also no benefit. So I'm fine with this as is.

PR #622, opened 2025-04-11
docs: add TOC to readme (fixes #333)

### LINKED ISSUE #333: docs: (IAC-1290) add a ToC in the main README file
It is minor, but I think it would improve the documentation readability to have a clickable table of content in the main README file (https://github.com/sassoftware/viya4-deployment/blob/main/README.md) with links to the distinct sections and subsections (Overview, pre-requisites, etc...). Also would allow to reference specific parts of the doc with anchor links.

sas-grtoma:


saschjmil (approved):
Looks good to me.

iadomi:
Are we only adding TOC for the main readme?  I'm guessing the other .md either don't need it or already have one?

dhoucgitter (approved):

sas-grtoma:
@iadomi Not sure - we could always go back and add it to more .md files. Do you think we should?

iadomi:
> @iadomi Not sure - we could always go back and add it to more .md files. Do you think we should?

I think we're okay for now.  The original ask from it https://github.com/sassoftware/viya4-deployment/issues/333 seem to suggest it was only a problem for the main readme and not the .md's in the docs folder.  Quickly looking over some of the other .md's, they are much short in length

Issue #634, opened 2025-05-08
Allow inject-sas-bases-overlays to support sas-bases/components


ajeffowens:
### Is your feature request related to a problem? Please describe.

The new inject-sas-bases-overlays feature is incredibly useful and welcomed.  However, it only supports refs to sas-bases/overlays directory.  

For a configuration such as sas-singlestore, where the user is directed to include this in their kustomization:

```
components:
- sas-bases/components/sas-singlestore
```

Dac fails to process the ref.  Would it be possible to allow references to components as well?  

### Describe the solution you'd like

Update this block of code:

```
              requiredPrefix = "sas-bases/overlays/"
              for entry in entries:
                if entry.startswith(requiredPrefix):
```

to allow for components refs.  

### Describe alternatives you've considered

Making a static copy of the kustomizations from that sas-bases directory following the existing doc:  https://github.com/sassoftware/viya4-deployment/blob/main/docs/user/SingleStore.md is possible, but seemingly could be greatly simplified.  

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

saschjmil:
Hey @ajeffowens. This is a great request. I'll create an internal ticket for it, but it might be a bit before we can get to it. If anyone else wants to handle it, they are welcome to open a PR.

ajeffowens:
https://github.com/sassoftware/viya4-deployment/pull/640

saschjmil:
@ajeffowens can this be closed since the PR has been merged?

PR #640, opened 2025-05-20
fix:  Expand categories supported by inject-sas-bases-overlays feature

ajeffowens:
This change simply narrows the scope of the `requiredPrefixes` var from `sas-bases/overlays` to just `sas-bases`, in order to support more categories of sas-bases refs for injection.  The enum already covers the possible keys.

### DIFF
--- a/roles/vdm/library/siteconfig_info.py
+++ b/roles/vdm/library/siteconfig_info.py
@@ -67,7 +67,7 @@ class siteConfig(object):
                 overlay = Overlay(blockName)
               except ValueError:
                 continue
-              requiredPrefix = "sas-bases/overlays/"
+              requiredPrefix = "sas-bases/"
               for entry in entries:
                 if entry.startswith(requiredPrefix):
                   self.add_overlays(overlay, entry)


ephesused (changes requested):
This looks good to me; thanks!

I have just one request, to adjust the corresponding documentation. Will you please edit the top-level `README.md`, [at this line](https://github.com/sassoftware/viya4-deployment/blob/c371f5d/README.md?plain=1#L256), changing:

    Base kustomization.yaml additions from sas-bases/overlays

to:

    Base kustomization.yaml additions from sas-bases/

Thanks again.

ephesused (approved):
LGTM

On our end we're looking into what's happening with the failed workflow.

saschjmil (approved):
Looks good to me. I believe the failing workflow is due to this PR coming from a fork. We will need to look into it, but it shouldn't stop this from being merged.

saschjmil:
Changing this from a bug to an enhancement. It has the `fix:` conventional commit but I feel like it falls under the enhancement umbrella. Especially since the related issue #634 has the enhancement label.

PR #659, opened 2025-07-25
feat: add support for kubernetes 1.33 (PSCLOUD-76)

abhi44706:
### Changes
- Updated default kubectl version to 1.32.6 in Dockerfile and docs
- updated cert-manager chart version from 1.16.2 to 1.18.2
- Adjusted ingress-nginx chart version logic to default to 4.13.0 for K8s >= 1.31.x
- Updated cluster-autoscaler chart version to 9.48.0 for compatibility with newer K8s versions
- added fix to handle UndefinedMarker errors in Ansible Jinja templates

### Tests
|Scenario|DaC use method|Provider|K8s version|metrics-server app_version|cluster-autoscaler app_version|Order|Cadence|Postgres| Notes|
|-|-|-|-|-|-|-|-|-|-|
|1|local ansible|AWS|1.31|0.7.2|1.32.0|*|stable:2025.06| Crunchy |Successful baseline install of metrics-server and cluster-autoscaler at recorded app_version, successful Viya install with all pods Running and stable|
|2|docker container|AWS|1.32|0.7.2|1.32.0|*|stable:2025.06| Crunchy |Successful baseline install of metrics-server and cluster-autoscaler at the indicated app_versions, Successful Viya deployment, all pods Running. Successful login to Viya app as viya_admin. Successful creation of SAS Studio context in SAS Studio app. |
|3|docker container|Azure|1.33|n/a|n/a|*| stable:2025.06| Crunchy/External |Verified successful deployment operator deployment with kubectl 1.32.6 version override. Successful login to Viya as viya_admin to landing page. Successful creation of SAS Studio context in SAS Studio app.|
|4|docker container|GCP|1.32|n/a|n/a|*|stable:2025.03| Crunchy/External |Verified successful deploy command deployment. Successful login to Viya as viya_admin to landing page. Successful creation of SAS Studio context in SAS Studio app.|


docs/CONFIG-VARS.md
```diff
@@ -414,7 +414,7 @@ By default, two block storage StorageClasses are created using the driver, one f
...
 | INGRESS_NGINX_CHART_NAME | NGINX Ingress Helm chart name | string | ingress-nginx | false | | baseline |
-| INGRESS_NGINX_CHART_VERSION | NGINX Ingress Helm chart version | string | "" | false | If left as "" (empty string), version `4.12.1` is used for Kubernetes clusters whose version is >= 1.28.X, for Kubernetes clusters whose version is <= 1.27.X you must set this variable to avoid errors. See [Supported Versions table](https://github.com/kubernetes/ingress-nginx/?tab=readme-ov-file#supported-versions-table) for the supported versions list. | baseline |
+| INGRESS_NGINX_CHART_VERSION | NGINX Ingress Helm chart version | string | "" | false | If left as "" (empty string), version `4.13.0` is used for Kubernetes clusters whose version is >= 1.31.X, for Kubernetes clusters whose version is <= 1.33.X you must set this variable to avoid errors. See [Supported Versions table](https://github.com/kubernetes/ingress-nginx/?tab=readme-ov-file#supported-versions-table) for the supported versions list. | baseline |
```
saschjmil, line 417: Why does this change the previous behavior? I feel like everything after the `4.13.0` shouldn't change.

roles/baseline/defaults/main.yml
```diff
@@ -37,8 +37,8 @@ ingressVersions:
   k8sMinorVersion:
     value: 28
     api:
-      chartVersion: 4.12.1
-      chartVersionMin: 4.0.17
+      chartVersion: 4.13.0
+      chartVersionMin: 4.12.0
```
saschjmil, line 41: This seems like a big jump. Why are we preventing users from using the 4.0.17+ charts?

saschjmil (changes requested):
This is looking good. The ansible-lint linter is failing so that will need to be addressed. Could you add your testing scenarios to the PR description? Something like #621.

Also, I'm concerned about the `INGRESS_NGINX_CHART_VERSION` changes.

docs/CONFIG-VARS.md
dhoucgitter: Given the support range for ingress-nginx 4.13, I'd expect the suggested values in the updated entry:
```suggestion
| INGRESS_NGINX_CHART_VERSION | NGINX Ingress Helm chart version | string | "" | false | If left as "" (empty string), version `4.13.0` is used for Kubernetes clusters whose version is >= 1.29.X, for Kubernetes clusters whose version is <= 1.28.X you must set this variable to avoid errors. See [Supported Versions table](https://github.com/kubernetes/ingress-nginx/?tab=readme-ov-file#supported-versions-table) for the supported versions list. | baseline |
```
since you'd want to use 4.13.0 for k8s 1.29 or newer clusters and you'd need the customer to specify which ingress-nginx version to use for anything at k8s 1.28 or older. Also, the intent is for no versions to exist between the last two >= and <= values in the sentence.

Make sure to include tests around the k8s boundary such as k8s 1.28 and 1.29 to make sure you get the expected behavior from DaC. If the changed values match my suggestion, then deploying to a k8s 1.28 cluster with an empty ingress-nginx value should result in an error asking for it to be set.

dhoucgitter: Given the support range for ingress-nginx 4.13, I'd expect the suggested values in the updated entry:
```suggestion
| INGRESS_NGINX_CHART_VERSION | NGINX Ingress Helm chart version | string | "" | false | If left as "" (empty string), version `4.13.0` is used for Kubernetes clusters whose version is >= 1.29.X, for Kubernetes clusters whose version is <= 1.28.X you must set this variable to avoid errors. See [Supported Versions table](https://github.com/kubernetes/ingress-nginx/?tab=readme-ov-file#supported-versions-table) for the supported versions list. | baseline |
```
since you'd want to use 4.13.0 for k8s 1.29 or newer clusters and you'd need the customer to specify which ingress-nginx version to use for anything at k8s 1.28 or older. Also, the intent is for no versions to exist between the last two >= and <= values in the sentence.

I'd also include tests around the k8s boundary such as k8s 1.28 and 1.29 to make sure you get the expected behavior. If the changed values match my suggestion, then deploying to a k8s 1.28 cluster with an empty ingress-nginx value should result in an error asking for it to be set.

dhoucgitter: Given the support range for ingress-nginx 4.13, I'd expect the suggested values in the updated entry:
```suggestion
| INGRESS_NGINX_CHART_VERSION | NGINX Ingress Helm chart version | string | "" | false | If left as "" (empty string), version `4.13.0` is used for Kubernetes clusters whose version is >= 1.29.X, for Kubernetes clusters whose version is <= 1.28.X you must set this variable to avoid errors. See [Supported Versions table](https://github.com/kubernetes/ingress-nginx/?tab=readme-ov-file#supported-versions-table) for the supported versions list. | baseline |
```
since you'd want to use 4.13.0 for k8s 1.29 or newer clusters and you'd need the customer to specify which ingress-nginx version to use for anything at k8s 1.28 or older. Also, the intent is for no versions to exist between the last two >= and <= values in the sentence.

Make sure to include tests around the k8s boundary such as k8s 1.28 and 1.29 to make sure you get the expected behavior from DaC. If the changed values match my suggestion, then deploying to a k8s 1.28 cluster with an empty ingress-nginx value should result in an error asking for it to be set.

dhoucgitter (changes requested):
See my version suggestions surrounding the CONFIG-VARS.md version values 
since you'll want to use 4.13.0 for k8s 1.29 or newer clusters and you'd need the customer to specify which ingress-nginx version to use for anything at k8s 1.28 or older. Also, the intent is for no versions to exist between the last two >= and <= values in the sentence.

Also wondering what made the chartVersionMin change from 4.0.17 to 4.12.0 necessary?

lohitdave (approved):

niraj2347 (approved):

PR #663, opened 2025-08-11
feat: set ingress-nginx chart version dynamically based on cadence version (PSCLOUD-99, PSCLOUD-158)

abhi44706:
set ingress-nginx chart version dynamically based on cadence version

- Added conditional logic to set INGRESS_NGINX_CHART_VERSION to 4.12.1 when
  V4_CFG_CADENCE_VERSION <= 2025.07, otherwise default to 4.13.0.
- Ensures correct chart version is used during Helm deployment based on cadence schedule.

### DIFF
--- a/roles/baseline/tasks/ingress-nginx.yaml
+++ b/roles/baseline/tasks/ingress-nginx.yaml
@@ -27,6 +27,10 @@
     - install
     - update
   block:
+    # Set the chart version based on cadence version
+    - name: Set ingress-nginx chart version based on cadence
+      set_fact:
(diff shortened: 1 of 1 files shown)


hiteshzope (approved):

saschjmil:
@abhi44706 @hiteshzope This is changing the default behavior around ingress-nginx chart versions. Previously, if the user didn't specify a chart version, it was chosen based on the k8s version. It was intentionally not set if K8s version was <= to the documented value: https://github.com/sassoftware/viya4-deployment/blob/main/docs/CONFIG-VARS.md#ingress-nginx

Now, it will always be set based on the cadence version.

PR #775, opened 2026-08-31
fix: add IPv6 Pod CIDR to LOADBALANCER_SOURCE_RANGES

joboon:


joboon:
@abhikumar2204 this is what I've staged for fixes to FNDTRG-1896. Take a look and let me know if you have any thoughts. Also, am I merging into the right branch?

joboon:
Also, I did test this with a fresh viya4-iac-aws deployment and viya4-deployment baseline and validated that the right VPC CIDR was added to the contour-envoy LoadBalancer service's loadBalancerSourceRanges list. I also validated that I could hit the Contour IPv6 address from within a Pod hosted in EKS.

